PR #19, opened 2020-02-11
Add GC for the ethbridge and add retries for the ethrelay

evgenykuzyakov:


k06a:
Is it WIP? I see `break` in loop in tests and some `env::log` items.

evgenykuzyakov:
> Is it WIP? I see `break` in loop in tests and some `env::log` items.

No. Logs are actually helpful to see inner state of GC and what's the result when submitting a header.
`break` is for retries, because sometimes calls fail and we need to retry them.

evgenykuzyakov:
Here is how it looks from the `index.js` logs:
https://gist.github.com/evgenykuzyakov/37f760e395dba6d11ff9538a21ea0bf9

PR #23, opened 2020-02-19
Add matrix for rust and os in CI

k06a:


nearmax:
@ailisp @k06a needs your help to finish this PR. @k06a could you provide more context?

ailisp:
I took a quick look. On ubuntu system builtin repo doesn't have node 10, so you need to install this way:
https://github.com/nodesource/distributions/blob/master/README.md#installation-instructions

On mac, it fails because there's no yarn. Install yarn after install node@10 will work

k06a:
@ailisp please help with NodeJS@8 or NodeJS@10 installation, it should be accessible by `node` command. Hope `yarn` will work also :)

PR #31, opened 2020-04-22
Rainbowup. First version

nearmax:
Missing features:
* Currently does not start nearrelay;
* We need to integrate ethprover tests;
* Various DevOps improvements, like dockerization of the services.


nearbridge/NearBridge.full.sol
```diff
@@ -370,6 +370,7 @@ library Borsh {
     function decodeBytes32(Data memory data) internal pure shift(data, 32) returns(byte[32] memory value) {
         bytes memory raw = data.raw;
         uint256 offset = data.offset;
+        // solium-disable-next-line security/no-inline-assembly
```
k06a, line 373: Add these changes to the original file in `contracts` directory and produce this full file with `yarn run dist` or `npm run dist`.

nearmax: These changes were autogenerated by `yarn run dist`

PR #32, opened 2020-05-03
Refactor bridge into services with shared functionality

nearmax:
At the end the folder structure will be the following:
* `near-contracts` -- folder with Near contracts (no js code);
* `eth-contracts` -- folder with Solidity contracts;
* `rainbowup`;
* `services` -- one node.js app that can do various things depending on the passed command:
  * Initialize necessary contracts both on Near and Ethereum;
  * Start production services, like relayers;
  * Start testing services, like the one that hammers ethprover.


nearmax:
Merging for now the piece with EthRelayer

PR #52, opened 2020-05-23
Fix web3 stuck

ailisp:
For any web3 ws connection, it needs manually close otherwise nodejs will wait it close before exit and stuck

### DIFF
--- a/environment/lib/eth-prover-tester/index.js
+++ b/environment/lib/eth-prover-tester/index.js
@@ -1,9 +1,9 @@
 const utils = require('ethereumjs-util');
 const Web3 = require('web3');
 const BN = require('bn.js');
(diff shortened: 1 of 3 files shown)


environment/lib/eth-prover-tester/index.js
```diff
@@ -35,7 +35,8 @@ class EthProverTester {
         let extractor = new EthProofExtractor();
         extractor.initialize(this.ethNodeURL);
         let test_num = 0;
-        for (let currBlock = firstBlock; firstBlock <= lastBlock; firstBlock++) {
+
+        for (let currBlock = firstBlock; currBlock <= lastBlock; currBlock++) {
```
ailisp, line 39: This looks like typo so i correct btw

Issue #47, opened 2020-05-15
Fix `next_bps_hash` computation on NearClient (NearBridge) side


k06a:
https://github.com/nearprotocol/nearcore/blob/79166fc254a4531ed047ae33d5229ac7983c6da8/chain/chain/src/chain.rs#L385-L394

PR #54, opened 2020-05-26
[WIP] Update NearBridge to the latest LightClient spec

k06a:


libs-sol/nearbridge/contracts/NearBridge.sol
```diff
@@ -123,32 +125,23 @@ contract NearBridge is Ownable {
...
-        //         nearBlock.next_bps.validatorStakes
-        //     ),
-        //     "NearBridge: Less than 2/3 voted by the block after next"
-        // );
+        // 4. approvals_after_next contain signatures that check out against the block producers for the epoch of the block
+        // 5. The signatures present in approvals_after_next correspond to more than 2/3 of the total stake
+        uint256 totalStake = 0;
+        uint256 votedFor = 0;
+        if (prev.bps_count > 0) {
+            require(nearBlock.next_bps.validatorStakes.length == prev.bps_count, "NearBridge: number of BPs should match number of approvals");
```
SkidanovAlex, line 133: As per spec, number of approvals can be smaller than the number of BPs

libs-sol/nearbridge/contracts/NearBridge.sol
```diff
@@ -123,32 +125,23 @@ contract NearBridge is Ownable {
...
+        }
+        for (uint i = 0; i < nearBlock.next_bps.validatorStakes.length; i++) {
+            totalStake = totalStake.add(
+                nearBlock.next_bps.validatorStakes[i].stake
+            );
+            if (!nearBlock.approvals_after_next[i].none) {
+                // Assume presented signatures are valid, but this could be challenged
+                votedFor = votedFor.add(nearBlock.next_bps.validatorStakes[i].stake);
```
SkidanovAlex, line 141: Seems like we are using block producers for the **next** epoch?

k06a:
Fixed #47

libs-sol/nearbridge/contracts/NearBridge.sol
k06a: Will fix that

libs-sol/nearbridge/contracts/NearBridge.sol
k06a: Some approvals are empty but lengths should match, isn't it?

SkidanovAlex: No, the length can be shorter, in which case it is assumed that the rest is `None`s.

It is not very noticeable in the spec, but it's there:

> approvals_after_next can contain fewer elements than next_bps in the previous epoch.

(https://github.com/nearprotocol/NEPs/blob/master/specs/ChainSpec/LightClient.md)

PR #61, opened 2020-05-30
Added prepare and clean commands

chefsale:


environment/commands/clean/index.js
```diff
@@ -1,7 +1,7 @@
+const {exec} = require('child_process');
+
 class CleanCommand {
-  static async execute() {
-    console.log("TODO (chefsale): Implement CleanCommand");
-  }
+  static execute() { exec('rm -rf ~/.rainbowup && pm2 kill'); }
```
nearmax, line 4: I cannot find where `pm2` is used in our codebase, I could only find its declaration in `package.json`.

nearmax:
FYI This PR had two issues:
* Argument parsing in prepare.sh did not work, had to remove it;
* prepare.sh was assuming that something did not create `~/.rainbowup` directory before, but `RainbowupArgParser.__init__` does exactly that.

chefsale:
Thanks for fixing that, not sure about the arg parsing what went wrong. The shell script I believe should still check if the rainbowup folder a d the argparser should crate the folder. Its bad on my side I didn't test that it works in the python code, but only node. Thanks for the heads up, we can discuss this next week.

chefsale:
So, I added this diff: 
https://github.com/near/rainbow-bridge/commit/0d7cc373cc4c28879e364f332ff901635716b402.

Seperated the node.js from the python script because of the different behaviour, tested it out works.

The new flow I added for node.js is that if you specify your local folder and do prepare, you can do it only once. If you want to do it again you have to clean.

As well the other tihng I enforced is that all the folders live under ~/.rainbowup, I plan to add a configuration file in which you would be able to specify another one, so all the commands will use the same home instead of you having to specify it everytime which is kind of a weird thing to do.

environment/commands/clean/index.js
ailisp: I think once nearbridge is started with pm2 this would kill it. If it's not it's a no-op so it's okay.

PR #67, opened 2020-06-05
Add sequence diagram of the eth2near transfer

nearmax:


docs/README.md
```diff
@@ -0,0 +1,10 @@
+# Rainbow Bridge
+
+This documentation is about design, usage, maintenance, and testing of the NEAR-ETH bridge.
+
+* [Overview of the components](./components.md)
```
alexfmpe, line 5: I was looking for docs and hit a broken link here.
This file doesn't seem to exist anywhere in the repo?

nearmax: We forgot to add it. Thanks for noticing.

PR #86, opened 2020-06-11
verify dumped eth headers and proofs in ci

ailisp:
Remove all other ci for now, as most of them has been moved. will add them back in this PR as they pass. Re-org tests as single script, remove env var: it was N env var  x N task (but (N-1)*N of them is no-op due to the guard if), make it N task now.
benefit of moving them to scripts: become one liner in travis ci, and the ci/*.sh is also locally, so it's easier to debug.

Note: eth-headers has enough block to be a 20 minutes test locally. but when batch fetching eth prpofs, hit infura rate limit:
```
(node:38546) UnhandledPromiseRejectionWarning: Error: Returned error: daily request count exceeded, request rate limited
```
might need fetch more later


ailisp:
Above two build are identical with different name, one in travis-ci.org on travis-ci.com, but one fail one pass 🤔 (also it shouldn't build twice)

nearmax:
Thank you @ailisp !

PR #87, opened 2020-06-12
adding ci tests back

ailisp:
Run all existing tests on buildkite ci (travis is too slow and unreliable and not debuggable for those test that need spin up a ganache), fix lint


ailisp:
Travis is too slow and hard to debug for integration test (i can't ssh to see why ganache is not launched in near bridge test), will switch to buildkite.

ailisp:
buildkite is even slower 😂 but hope it better after tweak aws cloudformation, and after debug merge several yarn into one would be fast

ailisp:
@nearmax I think let's merge this PR before others to have a working CI for other PRs

PR #90, opened 2020-06-15
Dockerization of the rainbow services

chefsale:
Both prod and dev setup


Dockerfile
```diff
@@ -0,0 +1,33 @@
+FROM ubuntu:18.04
+
+COPY . /usr/src/
```
ailisp, line 3: This step is better after apt-get and install rust, so layers can be can cached

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+ENV RUSTUP_HOME=/usr/local/rustup \
+    CARGO_HOME=/usr/local/cargo \
+    PATH=/usr/local/cargo/bin:$PATH
+
+RUN curl https://sh.rustup.rs -sSf | \
+    sh -s -- -y --no-modify-path --default-toolchain nightly-2020-05-15
+
+
+COPY config* ~/.rainbowup
```
ailisp, line 30: I think config should be mount as volume, instead of baked into docker image. so people don't rebuild image for config changes

README.md
```diff
@@ -5,12 +5,26 @@
...
-  
+
   <p>
     <a href="https://travis-ci.com/near/rainbow-bridge"><img src="https://travis-ci.com/near/rainbow-bridge.svg?branch=master" alt="Travis Build" /></a>
   </p>
 </div>
 
+## Configs and flags
+
+There are three layers of configuration, overriding each other with highest priorities first:
```
ailisp, line 16: This is great! very clear

README.md
```diff
@@ -31,3 +45,43 @@ Then run
...
 ```
 You must observe blocks being submitted.
+
+Docker:
+
+## Currently we have the following docker options:
+
+1. Rainbow Docker image containing rainbowup ready for running
+	- run the rainbowup docker image with a custom command
+2. A development docker compose setup (docker-compose-dev.yml)
```
ailisp, line 55: would be better to name it just docker-compose.yml, because local development happens more often than prod. and do every docker-compose command with `-f docker-compose-dev.yml` is lengthy

docker-compose-dev.yml
```diff
@@ -0,0 +1,30 @@
+version: "3.3"
+services:
+  near-node:
+    image: rainbowup-image
+    build: .
+    network_mode: host
+    command: bash -c "node index.js start near-node; node index.js start near-node && while true; do sleep 30; done;"
+  ganache:
+    image: rainbowup-image
```
ailisp, line 9: It's better to specify dockerimage, so it builds based on a local dockerfile:
https://docs.docker.com/compose/compose-file/#dockerfile

environment/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.14.2-buster as build
+FROM ubuntu:18.04
```
ailisp, line 1: This change is wrong, ethhashproof need golang to build. Also there must be `as` since dockerfile cannot have two plain `FROM`

Dockerfile
chefsale: This files is needed in that repo which gets used inside the image, even if we make this a volume we would need manual copy for the config inside the container. We can discuss this in more detail.

Dockerfile
chefsale: Changed. Good catch.

README.md
chefsale: I agree, on that. I just wanted to be more explicit, so people don't get confused by thinking the docker-compose.yml is the production one.

Dockerfile
ailisp: No, if you mount config* as volume ~/.rainbowup, from container's perspective config* file is a file inside container, you don't need to copy that.

docker-compose-dev.yml
chefsale: This still should build from the local image. This should just ensure we don't rebuild it several times, but once.

README.md
ailisp: Makes sense, but
docker-compose.yml
docker-compose-prod.yml

would be list together, due to their filename, on github-ui or a local file browser, so it's unlikely people would ignore the prod one though. And also, we'll have documentation on running prod, people need to read that anyway for a production deployment, there we can explicitly mention docker-compose-prod.yml

environment/Dockerfile
chefsale: That is my mistake but not sure even if we need this anymore. I was thinking about deleting this file all together because we don't have the ethhashproof command implemented yet, it was implemented in the python version if I recall correctly. Currently this is not used. So maybe we should delete it and add the golang once we have the nodejs implementation done. cc: @nearmax your opinion?

docker-compose-dev.yml
ailisp: No, if you `docker-compose build` it's smart enough to build it once

environment/Dockerfile
ailisp: I think this file can be deleted, however our major dockerfile need to strip more "development dependencies" as a production docker image, like what we did in this dockerfile:
```
FROM golang:... as build-ethhashproof
# build ethhashproof
FROM rust:... as build-rust-contracts
# build a few wasm files
FROM node:14
# The entry point nodejs app
# build sol contracts 
# copy built ethhashproof binary into image
# copy built rust contracts into image
```

Dockerfile
chefsale: Cool, that's true. We need to change the prepare script to copy the config file to ~/.rainbowup in that case first and than we can just mount it there. That makes sense. Do you mind if I do it as a follow up issue?

README.md
chefsale: Ok, will change it to docker-compose.yaml than. Makes sense to me :)

docker-compose-dev.yml
ailisp: Also `docker-compose build` would be a bit more common practice than `docker build . -t rainbowup-image`

chefsale: so, you mean if I just specify build: Dockerfile everywhere that should work?

Dockerfile
ailisp: Sure!

chefsale: Created an issue for this: https://github.com/near/rainbow-bridge/issues/92.

docker-compose-dev.yml
ailisp: Yeah. we used this way in explorer: https://github.com/near/near-explorer/blob/master/docker-compose.yml

chefsale: Done.

environment/Dockerfile
chefsale: Deleted the file.

ailisp:
update branch to avoid ci stuck

chefsale:
Ping @nearmax for review. :)

Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM ubuntu:18.04
```
nearmax, line 1: We could use https://github.com/rust-lang/docker-rust here.

docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+    command: bash -c "node index.js start near-node; node index.js start near-node && while true; do sleep 30; done;"
+  ganache:
+    build: .
+    network_mode: host
+    command: >
+      bash -c "
+        while ! nc -z localhost 3030; do
+          sleep 1
+        done
+        node index.js start ganache --daemon false
```
nearmax, line 15: This issue unfortunately breaks it: https://github.com/near/rainbow-bridge/issues/96

Dockerfile
ailisp: I think instead of using rust, we should a nodejs image, because
- we have rust and golang, but they're all dev depencies
- resulting service need nodejs, wasm files, ethhashproof binary, they can all run in a nodejs based image

ailisp:
Looks there's a lint fails on CI, would probably help to add to pre commit hooks

nearmax:
> Looks there's a lint fails on CI, would probably help to add to pre commit hooks

@ailisp That would be great. Could you please add it?

ailisp:
> @ailisp That would be great. Could you please add it?

This seems easier for us to install a plugin on your editor side, for example, vscode you can
 choose: https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint
and set:
```
"editor.codeActionsOnSave": {
        "source.fixAll": true
    }
```

clion:
https://www.jetbrains.com/help/clion/eslint.html# 
> To fix all detected problems automatically when your project files are saved, select the Run eslint --fix on save checkbox. 


There's an option to add as pre commit hook, but you still need to either install this: (https://pre-commit.com/), or require manually copy a pre-commit script into your ~/.git/hooks, i feel it's not as convenient as just install above eslint plugins into editor

nearmax:
> > @ailisp That would be great. Could you please add it?
> 
> This seems easier for us to install a plugin on your editor side, for example, vscode you can
> choose: https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint
> and set:
> 
> ```
> "editor.codeActionsOnSave": {
>         "source.fixAll": true
>     }
> ```
> 
> clion:
> https://www.jetbrains.com/help/clion/eslint.html#
> 
> > To fix all detected problems automatically when your project files are saved, select the Run eslint --fix on save checkbox.
> 
> There's an option to add as pre commit hook, but you still need to either install this: (https://pre-commit.com/), or require manually copy a pre-commit script into your ~/.git/hooks, (since with either approach you need a one time manually setup anyway) i feel it's not as convenient as just install above eslint plugins into editor


Okay, thanks.

Dockerfile
chefsale: Personally I think we should stick to writing our own docker files, using same operating system. In this case if we just stick to ubuntu:18.04 everywhere gives us full power of what do we want to install and how (same package manager, versioning, commands, shell, kernel, etc.) . As well, it gets easier to use the docker images itself, as they are all ubuntu. We could though create our own nodejs, go and rust default images and build on top of them.As these docker images still use Debian which is fine, but I think in our case we should strive to get our base images built in-house.

ailisp: I'm fine to stick on ubuntu, but for a production quality image, only install nodejs. Golang and rust are only _development dependencies_ and should be avoided in result image.

ailisp: Refer to, for example: https://github.com/nearprotocol/nearcore/blob/master/Dockerfile#L46
on how we avoid rust and compilation intermediates into result file (which result in only ~10% in size compare to have rust etc. included)

chefsale: Yep, currently this image is also used for development, that's why we are bundling both ganache and local neard image.

docker-compose.yml
chefsale: As commented on: https://github.com/near/rainbow-bridge/issues/96. Adding the command as is hardcoded for now with an installation of npm -g install ganache-cli in the image
```
ganache-cli --port 9545 --blockTime 12 --gasLimit 10000000 --account='0x2bdd21761a483f71054e14    f5b827213567971c676928d9a1808cbfa4b7501200,10000000000000000000000000000'
```

PR #85, opened 2020-06-11
Added initial configstore for rainbowup

chefsale:
Test Plan:
```

sandi@sandi-ThinkPad-X1-Carbon-7th ~/w/n/r/environment (commands/init-eth-contracts)> node index.js init-eth-contracts --near-network-id test --near-node-url localhost
The current Rainbow configuration is:
{ nearNetworkId: 'test',
  nearNodeUrl: 'localhost',
  ethNodeUrl: '' }
sandi@sandi-ThinkPad-X1-Carbon-7th ~/w/n/r/environment (commands/init-eth-contracts)> cat ~/.rainbowup/config.json
{
	"nearNetworkId": "test",
	"nearNodeUrl": "localhost",
	"ethNodeUrl": ""
}⏎                                                                                                                                                                                                                  sandi@sandi-ThinkPad-X1-Carbon-7th ~/w/n/r/environment (commands/init-eth-contracts)>

```

PR #93, opened 2020-06-16
Move tests, upgrade near-sdk, borsh

evgenykuzyakov:
Required for importing nearcore runtime standalone and write integration test


ailisp:
cancel build as #85 introduce a regression cause ci stuck forever

libs-rs/rust-toolchain
```diff
@@ -0,0 +1 @@
+nightly-2020-05-15
```
nearmax, line 1: Our contracts should be compilable with stable. What would be the reason to use nightly here?

ailisp: For intention to run cross contract test:
```
near-runtime-standalone = { git = "https://github.com/nearprotocol/nearcore.git" }
```
and nearcore need nightly to build. Maybe we should
- specify stable-xxx in rust-toolchain to produce determinstic build on contracts
- use script to switch to nightly to build cross contract test

ailisp: Oh actually @evgenykuzyakov added
```
RUSTFLAGS='-C link-arg=-s' cargo +stable build --target wasm32-unknown-unknown --release
```
in build.sh, this is how we did in core-contracts, so contracts are built with stable. but problem is we can't specify a determined rust version here

ailisp:
Ping @nearmax ptal :)

PR #97, opened 2020-06-17
Clippy and other fixes

chain-police:
- clippy pass
- fix tests
- formatting
- overflow-checks = true


Kouprin:
PR bankruptcy rule

PR #98, opened 2020-06-18
Cross contract test

ailisp:
Cross contract test, test a func in eth-prover that call into eth-client


libs-rs/eth-prover/tests/spec.rs
```diff
@@ -0,0 +1,32 @@
...
+    };
+    near.function_call(&mut r, "eth-client", "add_block_header", &add_block_header_args.try_to_vec().unwrap(), ntoy(0)).unwrap();
+
+    testing_env!(get_context(vec![], false)); // For use rlp::decode, eth_types::near_keccak256
+    let header: BlockHeader = rlp::decode(block.header().as_slice()).unwrap();
+    let assert_ethbridge_hash_args = AssertEthbridgeHashArgs {
+        block_number: header.number,
+        expected_block_hash: header.hash.unwrap(),
+    };
+    println!("{:?}", near.function_call(&mut r, "eth-prover", "assert_ethbridge_hash", &assert_ethbridge_hash_args.try_to_vec().unwrap(),ntoy(0)).unwrap());
```
ailisp, line 30: Question, this function supposed to gives SuccessValue(`true`) or SuccessValue(`false`), but it gives:
```
ExecutionOutcome { logs: [], receipt_ids: [], burnt_gas: 2592949332078, status: SuccessValue(``) }
```
Where is wrong? ping @nearmax @fckt for help. Will continue fix tomorrow

nearmax: CC @willemneal

ailisp: puzzle resolved by @evgenykuzyakov the return was `\x00` or `\x01` and not show up in my terminal, but it can be compared in rust :)

libs-rs/build_all.sh
```diff
@@ -5,6 +5,9 @@ set -e
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 
-$DIR/eth-client/build.sh
-$DIR/eth-prover/build.sh
-$DIR/fungible-token/build.sh
+cd $DIR/eth-client
+./build.sh
```
ailisp, line 9: Found cargo doesn't actually build if don't cd, and the later cp target/.../eth-prover.wasm ../ still copy, so that build_all.sh was not very helpful

PR #116, opened 2020-06-23
nearlib -> near-api-js and tsc check

### LINKED ISSUE #110: Upgrade nearlib to the most recent near-api-js
Currently we are using nearlib, but we should be using the most recent `near-api-js`.
Upgrading might require refactoring some code.

willemneal:
Now `yarn check` will now ensure that the types used are correct.  However, it does not perform any more checks.

fixes #110


environment/commands/init-near-fun-token.js
```diff
@@ -20,7 +20,7 @@ class InitNEARFunToken {
         const nearNodeUrl = RainbowConfig.getParam('near-node-url');
         const nearNetworkId = RainbowConfig.getParam('near-network-id');
 
-        const tokenPk = nearlib.KeyPair.fromString(tokenSk).publicKey;
+        const tokenPk = nearlib.KeyPair.fromString(tokenSk).getPublicKey();
```
willemneal, line 23: This seems the only real API change, at least given near-api-js types.

willemneal:
@nearmax How do I go about testing this?  Is CI sufficient?

PR #119, opened 2020-06-24
end to end test

ailisp:
Works locally, but even locally it stuck forever after successful transfer-eth-erc20-to-near. Need to figure out a few things before merge:
- [x] install golang in buildkite image
- [x] rainbowup eth-relay was always use ./enviroment/vendor/ethhashproof, but only ~/.rainbow/bridge/enviroment/vendor/ethhashproof is built (fixed only in ci, probably want it nicer in init/prepare scripts)
- [x] as said transfer-eth-erc20-to-near hangs


ci/e2e.sh
```diff
@@ -0,0 +1,28 @@
...
+trap 'pkill -15 -P $GANACHE_PID' 0
+node index.js init-near-contracts
+node index.js init-eth-ed25519
+node index.js init-near2eth-client
+node index.js init-eth-erc20
+node index.js init-eth-locker
+node index.js init-near-fun-token
+node index.js start eth-relay
+node index.js start near-relay
+node index.js transfer-eth-erc20-to-near --amount 1 --eth-sender-sk 0x2bdd21761a483f71054e14f5b827213567971c676928d9a1808cbfa4b7501200 --near-receiver-account alice.test.near --near-master-account nearfuntoken
```
nearmax, line 28: @ailisp Currently transfer command hangs at the end of the execution. How do we terminate this test?

nearmax:
Either web3 or near-api-js are holding the socket or something https://github.com/near/rainbow-bridge/issues/112

environment/lib/config.js
```diff
@@ -64,8 +64,12 @@ class RainbowConfig {
     static getParam (name) {
         const res = this.maybeGetParam(name);
         if (res === null) {
-            console.error(`Parameter ${name} must be specified.`);
-            process.exit(1);
+            if (!this.optionalValues.includes(name)) {
```
nearmax, line 67: This seems unnecessary. The code here already specifies that these arguments have default value as an empty string: https://github.com/near/rainbow-bridge/blob/b9e2f475fcd2e1250f3b36020511bdd531369385/environment/index.js#L91

ailisp: Without revert this change prepare gives error `Parameter bridge-src must be specified`, will check maybeGetParam

ailisp: I see the problem, `''` is treat as false in javascript, and so maybeGetParam returns false, will change maybeGetParam to check `undefined` to tell "parameter given (can be empty)"

nearmax: Thanks

PR #122, opened 2020-06-24
Fix poluted CI with old environment

chefsale:
Fix for  https://github.com/near/rainbow-bridge/issues/121


ailisp:
Looks cleanup logic is not fully works:

image

Locally i have this issue too, but rerun, cancel it works, but needs to be more robust to being used in CI

ailisp:
Buildkite sometimes reuse instance is indeed a problem, currently it does is: after a build finish, if there's no build in 60 seconds (configurable), it will terminate that instance. But if another build comes in in 60 seconds it reuse. There's an option to disconnect agent so an instance would not receive new job, and after 60s get deleted: 
search disconnect in https://buildkite.com/docs/agent/v3/configuration

Can you make the change on buildkite images? so we don't need worry about cleanup

chefsale:
I added the fix for the clean, as well. We could enable that option, but in that case we would need for new instances to get spinned up every time.

chefsale:
So, there is also the go issue: 
```
[2020-06-24T19:06:37Z] The Go install directory (/var/lib/buildkite-agent/.go) already exists. Exiting.
 ```

This still should fix the clean being broken. I will also disable the reusing of the machine for now.

ailisp:
> I added the fix for the clean, as well. We could enable that option, but in that case we would need for new instances to get spinned up every time.

We should enable that option, note https://buildkite.com/organizations/nearprotocol/agents we already delete and create new instances every time, and this gives us deterministic behavior to test from fresh made instances. But it's a pain that occasionally some instance get reused.

chefsale:
Yep, I'm changing the CloudFormation template for the instances now. cc: @ailisp

chefsale:
The minimum time that can be added is 1 second, so this will still happen but less than usually. So, in that case we still should make sure we have proper cleaning after jobs are terminated or before they are started. cc: @ailisp

ailisp:
> The minimum time that can be added is 1 second, so this will still happen but less than usually. So, in that case we still should make sure we have proper cleaning after jobs are terminated or before they are started. cc: @ailisp

No it won't happen if you enable `disconnect-after-job` in instance config in: https://buildkite.com/docs/agent/v3/configuration
Builkite agent daemon will immediately disconnect after job finish, then instance is deleted 60s (1s after your change) later, but within this 60s (1s) agent is shutdown, it won't accept new jobs.

chefsale:
Cool, will bake a new image with disconnect-after-job than :)

ailisp:
> [2020-06-24T19:06:37Z] The Go install directory (/var/lib/buildkite-agent/.go) already exists. Exiting.

This timestamp happens before my fix, and looks okay now:
image

And it won't be an issue anyway, after @chefsale enables disconnect-after-job

PR #123, opened 2020-06-25
dump and test near header

ailisp:
This one of near headers shown as pass, but checkBlockProducerSignatureInLastBlock fail and was commented: https://buildkite.com/nearprotocol/rainbow-bridge/builds/166

I'm still trying to figure out why checkBlockProducerSignatureInLastBlock return false


bowenwang1996:
We figured out that the issue is with block timestamp not encoded in string and exceeds the u53 limit in json. Here is the nearcore PR to address it https://github.com/nearprotocol/nearcore/pull/2912

ailisp:
@nearmax verify near dump test finally passed, with dump from new devnet deploy after nearprotocol/nearcore#2912 merged

nearmax:
Awesome. Thank you.

PR #124, opened 2020-06-25
Fun token extracts amount and recipient from the log entry

nearmax:


nearmax:
Only two tests are failing: e2e (which is not fixed yet) and test_verify_near_headers which is not implemented yet. Therefore I am merging this PR.

PR #125, opened 2020-06-25
Fun token should work only with hardcoded locker

nearmax:


nearmax:
Only verify near headers is failing on the CI , because the test was not merge into master. Merging this PR.

PR #131, opened 2020-06-26
configurable lock duration and eth amount in nearbridge contract

ailisp:
nearbridge contract  constructor now takes 3 arguments (add lock duration and lock eth amount)


ailisp:
^ nearbridge test actually pass, but run same test with coverage takes forever, didn't figure out why

libs-sol/nearbridge/contracts/NearBridge.sol
```diff
@@ -31,11 +31,13 @@ contract NearBridge is INearBridge {
         mapping(uint256 => BlockProducer) next_bps;
     }
 
-    uint256 constant public LOCK_ETH_AMOUNT = 1 ether;
-//    uint256 constant public LOCK_DURATION = 1 hours;
-    uint256 constant public LOCK_DURATION = 10 seconds;
+    uint256 constant public DEFAULT_LOCK_ETH_AMOUNT = 1 ether;
```
nearmax, line 34: `DEFAULT_*` are not used.

libs-sol/nearbridge/contracts/NearBridge.sol
```diff
@@ -31,11 +31,13 @@ contract NearBridge is INearBridge {
...
+//    uint256 constant public DEFAULT_LOCK_DURATION = 1 hours;
+    uint256 constant public DEFAULT_LOCK_DURATION = 10 seconds;
 
     bool public initialized;
+    uint256 public lock_eth_amount;
+    uint256 public lock_duration;
```
nearmax, line 40: Is it in seconds?

libs-sol/nearbridge/contracts/NearBridge.sol
```diff
@@ -31,11 +31,13 @@ contract NearBridge is INearBridge {
...
 
-    uint256 constant public LOCK_ETH_AMOUNT = 1 ether;
-//    uint256 constant public LOCK_DURATION = 1 hours;
-    uint256 constant public LOCK_DURATION = 10 seconds;
+    uint256 constant public DEFAULT_LOCK_ETH_AMOUNT = 1 ether;
+//    uint256 constant public DEFAULT_LOCK_DURATION = 1 hours;
+    uint256 constant public DEFAULT_LOCK_DURATION = 10 seconds;
 
     bool public initialized;
+    uint256 public lock_eth_amount;
```
nearmax, line 39: Is it in Eth or Wei?

ailisp: it's in wei, common practice in solidity is amount always in wei, and use ether is suffixed by the unit: `1 ether`, `ether` is then compiled to `* 1e18`

libs-sol/nearbridge/contracts/NearBridge.sol
ailisp: Yes, also common practice to be in second: https://solidity.readthedocs.io/en/v0.4.24/units-and-global-variables.html#time-units

libs-sol/nearbridge/contracts/NearBridge.sol
ailisp: removed

PR #134, opened 2020-07-01
Add assertions and transfer_eth_from_near to the e2e test

nearmax:


ci/e2e.sh
```diff
@@ -4,24 +4,43 @@
...
 ROOT_DIR=$CI_DIR/..
 
 cd $ROOT_DIR/environment
 yarn
+node index.js clean
 node index.js prepare
 node index.js start near-node
 node index.js start ganache
-./scripts/start_ganache.sh &
+./scripts/start_ganache.sh > /dev/null 2>&1 &
```
ailisp, line 16: Is this supposed to be `/tmp/ganache.out` and above one  `>/dev/null` ?

environment/lib/near2eth-relay.js
```diff
@@ -142,6 +142,7 @@ class Near2EthRelay {
     }
 
     async run () {
+        process.send('ready');
```
ailisp, line 145: what is this for?

nearmax: This signals to pm2 that the relayer has booted, done all initialization and is fully functional, see ` wait_ready: true,` configuration option.

ailisp: I see, does this need to be in the end of `run ()`?

ailisp:
Looks sometimes clearup still stucks: https://buildkite.com/nearprotocol/rainbow-bridge/builds/217#08e5c319-8f8b-4db6-9f38-6d98c8972e0f

environment/lib/near2eth-relay.js
nearmax: I should be between initialization and the main loop of the relay. If we put it at the end of run then it will never be emitted and pm2 will think this service is not "ready".

ailisp:
Recent test fails at coverage: https://buildkite.com/nearprotocol/rainbow-bridge/builds/221#cec10954-0884-45f8-af39-0ebce145bde2 I can't figure out why it would fail (it complain contract already initialized but coverage is a seperate ganache on 8555, so it shouldn't happen and cannot reproduce locally. I remove coverage for now.

ailisp:
@nearmax right now I rerun e2e test 6 times in master, result is stable and always pass (but master doesn't do tranfer-eth-from-near).  Also rerun a few times in this branch. still, transfer-eth-erc20-to-near always pass, but transfer-eth-from-near always fail, and of different reason:
```
[2020-07-02T19:05:39Z] (node:4214) UnhandledPromiseRejectionWarning: Error: Returned error: VM Exception while processing transaction: revert NearProver: block proof is not valid
```
```
[2020-07-02T23:25:06Z] (node:4251) UnhandledPromiseRejectionWarning: Error: [-32000] Server error: Other Error: block 8xViepvmUawa31iX3kQoBXmo7i7RrRwNwHiJpPLr3jLC is ahead of head block GXR1eFneDE8Wvtf3n8DKHhdscnWt2bpqr28HJqPabnY
```
Hopefully block proof not valid is because 1) not wait long enough or 2) fixed by the proof4's fix. but is ahead of head block looks like incorrect submitting or not waiting enough long in near-relay? will do further investigation there.

ailisp:
@nearmax This passed both locally and ci after a few PRs you did in weekend, please also review so we can merge it.

ailisp:
It's mostly reliable now, based on running here and a few times locally. A short period of e2e test like this is usually pass. Running this for long has more chance to encounter an unhandled error and therefore fail due to daemon die, some rpc fail without retry and missing some proof, they'll be address in #149. So merge this

PR #136, opened 2020-07-02
[WIP] Fix code in Eth2NearClient

nearmax:
* Fix garbage collection logic in the Eth2NearClient, not the cleaning criteria is more straightforward -- the headers past certain block number are GCed;
* Delegate decision on what is considered to a confirmed block outside the client (it should be eventually done by the fun token);

TODO:
* Update prover, fun token and relayers to work with the new client API.


libs-rs/eth-client/src/lib.rs
```diff
@@ -248,63 +260,59 @@ impl EthBridge {
...
+                    self.infos.remove(&hash);
+                }
+                self.all_header_hashes.remove(&header_number);
+                if header_number == 0 {
+                    break;
+                } else {
+                    header_number -= 1;
+                }
+            } else {
+                break;
```
ailisp, line 303: Is there any chance record_header fail, and then it add another header, and there's hole in gc?
```
header_number:10 header_number:11 (empty due to fail to record_header) header_number:12
```
Then before header_number:10 will never be cleaned

nearmax: Someone will have to resubmit it.

PR #137, opened 2020-07-03
dump verify near proofs

ailisp:
Similar as other dump verify tests. Added in CI


ailisp:
@nearmax verify near proofs test is passing, please review, thanks!

PR #141, opened 2020-07-06
Fix prover promise and header encoding. Fixes #114 . Fixes #140

### LINKED ISSUE #114: Contract fixes
Our contracts are missing the following important assertions:
- [x] The Near fungible token should accept the address of the corresponding locker contract when it is initialized and it should verify that the proof to the event that it receives corresponds to the locker contract that it was initialized with;
- [x] The Near fungible token should record hashes of the events that were used to mint and it should not allow the same proof to be used twice;
  - [ ] The code that sends the proofs to the fungible token should attach some tokens to account for the increase in storage;
- [ ] In Eth2NearProver contract the verification of the proof should be before the cross contract call to the client: https://github.com/near/rainbow-bridge/blob/c42defee097d2e4ac6d86777ae1c374e728402e1/libs-rs/eth-prover/src/lib.rs#L143 When `skip_bridge_call == false` we should be returning the result of the verification, and when `skip_bridge_call == true` the promise. Use `PromiseOrValue<bool>` for it, see: https://github.com/near/near-sdk-rs/blob/master/examples/cross-contract-high-level/src/lib.rs#L58

### LINKED ISSUE #140: Complex bug with Eth prover and Eth client
Update: The cause is found and a simple fix should be submitted by EOD July 6, 2020.

Currently, `transfer-eth-erc20-to-near` command waits until Eth2NearClient catches up with the necessary block before proceeding with minting. However, it does not wait for the block to be confirmed, instead it simply waits for it to appear on the canonical chain: https://github.com/near/rainbow-bridge/blob/master/environment/commands/transfer-eth-erc20-to-near.js#L146

This bug went uncaught by the tests because of another bug described in here: https://github.com/near/rainbow-bridge/issues/114 Specifically:
> In Eth2NearProver contract the verification of the proof should be before the cross contract call to the client: https://github.com/near/rainbow-bridge/blob/c42defee097d2e4ac6d86777ae1c374e728402e1/libs-rs/eth-prover/src/lib.rs#L143

Fortunately, it is an easy fix, however a naive fix breaks the tests. This happens because of another bug in https://github.com/zmitton/eth-object that we are using (and probably shouldn't). When tests are run with Ganache the difficulty is set to `0`, however `eth-object` incorrectly serializes Ethereum header with RLP, because `0` is a special case, see: https://ethereum.stackexchange.com/questions/67055/block-header-hash-verification Incorrect serialization results in incorrect header passed with the proof to the minting function, which causes it to crash.

nearmax:
Fixes #114 . Fixes #140


environment/lib/eth-proof-extractor/index.js
```diff
@@ -57,42 +57,36 @@ class EthProofExtractor {
...
+        // Correctly compose and encode the header.
+        const header = [
+            blockData.parentHash,
+            blockData.sha3Uncles,
+            blockData.miner,
+            blockData.stateRoot,
+            blockData.transactionsRoot,
+            blockData.receiptsRoot,
+            blockData.logsBloom,
+            blockData.difficulty == "0" ? "0x": web3.utils.toHex(blockData.difficulty),
```
ailisp, line 73: @nearmax I did some further investigation, this is correct, but in general, all other places that use number that could be 0, such as gasUsed, also have this problem, so I think we should add an additional pass to eth-object/eth-util-lite (will fork it), so it => [strings].map(toHex).map(if "0x0" else return =>"0x") => rlp encode will be correct

ailisp: Hmm looks i'm wrong, gasUsed is number, should be '0x0' when it's 0. But difficulty is a string, should be '0x' if it's '0' 🤔

PR #154, opened 2020-07-07
move pipeline definition in repo

ailisp:
It was follow nearcore's legacy branch pattern: master/beta/stable, that if pipeline were committed into repo, it diverge in master/beta/stable was a problem. but now rainbow-bridge only have every PR commit to master branch, and in future it probably will cut release just from master branch (same as current nearcore), so move pipeline file into repo gives benefit that PRs can modify pipeline file accordingly, such as rename enviroment -> rainbow_cli need to modify this pipeline file

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,0 +1,151 @@
+steps:
+    - command: |
+        source ~/.cargo/env
+        rustup default stable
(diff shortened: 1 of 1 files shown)


ailisp:
@chain-police FYI pipeline def is now in .buildkite/pipeline.yml

PR #156, opened 2020-07-09
retry txn on timeout error, near rpc error and nonce error

ailisp:
Fixe errors mentioned in https://github.com/near/rainbow-bridge/issues/142


ailisp:
The e2e tests passed for two times, due to the more retrying on timeout/connection/nonce, but it still fail for https://buildkite.com/nearprotocol/rainbow-bridge/builds/282#55a54435-5e1d-4463-a316-6444916a542e, stuck forever at: [2020-07-09T21:50:05Z] Eth2NearClient is currently at block 16. Waiting for block 11 to be confirmed. Sleeping for 10 sec.

This error happens, because even retry 10 times, submit block 2 eth relay still fails, and eth-relay crash and hangs forever (due to web3, it didn't exit), Fix by return to eth-relay's loop when this happens now

ailisp:
Got an new interesting failure because of the cruel retrying:
https://buildkite.com/nearprotocol/rainbow-bridge/builds/283#9ca9cc8d-ca24-4f12-9954-bbfdba6a4dd1

txn can timeout, but it eventually go through, cause nearTokenContractBorsh.mint being caused twice 🤔

environment/lib/borsh/index.js
```diff
@@ -193,27 +194,79 @@ function deserialize (schema, fieldType, buffer) {
...
+                continue;
+            } else {
+                // Indeed txn error, retry doesn't help
+                break;
+            }
+        } else {
+            // Still no result after a long time, resubmit txn
+            resendLast = true;
+            continue;
+        }
```
ailisp, line 266: can only come up with quite lengthy way to ensure txn successfully handled by nodes, hopes we can have such a reliable way in near-api-js

bowenwang1996: should be handled properly in 0.27.0

ailisp: Hmm just checked. Timeout is caught in 0.27.0, but these three cases are not:
- socket hangup
- retrying on correct nonce, in case submit with same key concurrently, or large nonce reach/process first, smaller nonce txn retry with correct nonce
- broadcast_txn wait seconds sometimes, cause socket hangup will not get a txn hash and also do not not txn is submit or not. Sometimes txn actually submitted. Then retry submit txn cause double txn send: https://buildkite.com/nearprotocol/rainbow-bridge/builds/283#9ca9cc8d-ca24-4f12-9954-bbfdba6a4dd1
I think expose broadcast_txn_async in near-api-js fix this (here i'm using jsonrpc to broadcast_txn_async), because this get response txnhash immediately, and we know it submitted, if we can get the response, then we just retry txn status (Just found this has exported) until get success or error.

Also Bowen mentioned even broadcast_txn_async and get a txn hash, due to our txn pool design it's every txn takes random long time to be picked up, it's possible some txn takes very long to going through, is this still the case? @bowenwang1996. Based on devnet load test, some takes even more than 100s, so in that case i resend txn with same nonce, and hopefully it has a larger chance to pickup, which has same nonce and won't be double processed.
cc @vgrichina

bowenwang1996: > it's possible some txn takes very long to going through

that might still happen, but very unlikely.

ailisp: Thanks. In near-relayer as long running process rare likelihood could add up (or when there's burst in node's load), i'll still handle this case

ailisp:
Seems robust now, https://buildkite.com/nearprotocol/rainbow-bridge/builds?branch=robust-error-handling
Five passes in a row

environment/lib/borsh/index.js
nearmax: We have pinned nearcore version in our prepare.sh script. Created an issue: https://github.com/near/rainbow-bridge/issues/161

ailisp: @nearmax Can you also approve this. It doesn't fail on contract call nonce, near&web3 timeout and rpc errors, check reinit a near contract in the past 6 runs. It fails once on deploy contract call nonce, but probably would solve in #161.

PR #157, opened 2020-07-09
make clean command more robust

### LINKED ISSUE #143: Handle killed pm2 daemon
Our `clean` command kills pm2 daemon, which means that sometimes when we do `start eth-relay` or other service they won't start, because we need to call `pm2 ping` first. We should fix it and call `ping` for every command that starts new daemon under pm2 daemon.

ailisp:
was originally intend to fix #143 , but it seems it works, see comments in #143. This fix the annoying error message showing up when the service was already cleaned


nearmax:
Thank you

PR #162, opened 2020-07-10
Update for new nearcore RPC

### LINKED ISSUE #161: Update nearcore to new version
Currently prepare.sh pins an old nearcore version, which causes us to have issues with nonces. We should bump it to a new version, but that would also require making changes to our header format in JS CLI and Solidity contracts to work with https://github.com/nearprotocol/nearcore/commit/c6cb9afc4ef01609cfeb678a377037304a22a2d5

nearmax:
Fixes #161


nearmax:
Deprecated.

PR #153, opened 2020-07-07
rename rainbowup to rainbow, enviroment to rainbow_cli for npm

ailisp:
Address #99

PR #167, opened 2020-07-13
Only rename rainbowup to rainbow for less confliction resolve

ailisp:
This replace #153


ailisp:
Merge as #153 was approved

PR #174, opened 2020-07-14
Parameter to run e2e test in nearcore ci

ailisp:
In rainbow-bridge, e2e.sh should take arg so it use local nearcore. In nearcore ci, it will checkout latest bridge, and run e2e.sh with current commit nearcore

### DIFF
--- a/environment/commands/prepare.js
+++ b/environment/commands/prepare.js
@@ -15,9 +15,9 @@ class PrepareCommand {
         for (const e in process.env) {
             env[e] = process.env[e];
         }
-        env.LOCAL_BRIDGE_SRC = path.resolve(RainbowConfig.getParam('bridge-src'));
-        env.LOCAL_CORE_SRC = path.resolve(RainbowConfig.getParam('core-src'));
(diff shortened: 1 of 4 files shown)


environment/commands/prepare.js
```diff
@@ -15,9 +15,9 @@ class PrepareCommand {
         for (const e in process.env) {
             env[e] = process.env[e];
         }
-        env.LOCAL_BRIDGE_SRC = RainbowConfig.getParam('bridge-src');
-        env.LOCAL_CORE_SRC = RainbowConfig.getParam('core-src');
-        env.LOCAL_NEARUP_SRC = RainbowConfig.getParam('nearup-src');
+        env.LOCAL_BRIDGE_SRC = RainbowConfig.getParam('bridge-src') && path.resolve(RainbowConfig.getParam('bridge-src'));
```
nearmax, line 18: Should it be just `path.resolve(RainbowConfig.getParam('bridge-src'))`?

ailisp: No that didn't work, because by default:
```
path.resolve(RainbowConfig.getParam('bridge-src'))
=> path.resolve('')
=> absolute path of rainbow-bridge dir
```

Issue #169, opened 2020-07-14
Validator signatures are broken on the most recent testnet


nearmax:
* Checkout https://github.com/near/rainbow-bridge/pull/164
* Uncomment https://github.com/near/rainbow-bridge/blob/e91e1245fdb4256d87ba056bafe26c32f827678e/libs-sol/nearbridge/test/NearBridge2.js#L41
* Go to `libs-sol/nearbridge` and run `yarn test`. Observe that the uncommented test is failing, while the old tests are passing. The old tests contain blocks from the old testnet and local networks, while the new tests contains blocks from the new testnet. Something broke the verification of the signatures.

ailisp:
We found the problem is with the data test is using. They're all from same epoch, and nearbridge doesn't have prev.next_bps, therefore unable to verify signature. If `initWithBlock` from epoch X, and `addLightClientBlock` from `epoch >= X+1` it can verify. In practice:

```
            prev.next_bps[signatureIndex].publicKey
```

is empty for the first epoch (because prev is not set until get block of second epoch), thus first epoch is not verifiable. To completely fix this prev.next_bps should be init to genesis validator sets.

nearmax:
@ailisp Could you please add this argument to the constructor and update the relay script?

PR #192, opened 2020-07-16
Fix 191 and 169

### LINKED ISSUE #191: Near2EthClient failed with NearBridge: number of BPs should match number of approvals
We were running bridge over night and suddenly transactions from `near-bridge` to `nearbridge` solidity contract started failing with: "NearBridge: number of BPs should match number of approvals". See the sequence of transactions and the moment it started failing here: https://ropsten.etherscan.io/txs?a=0x276d4d74dc14251c8d75ff4ae9175142e1c2254d&ps=100&p=21 The first failed transaction: https://ropsten.etherscan.io/tx/0x8fbc7ae814978a63602833cc7a36e91ca3880854df8f402f422ef226e322c83c

See the logs from the service here: https://drive.google.com/file/d/1nPx_CG62dG5mlR2t9GU-FdE2Mm2SumdQ/view?usp=sharing

Unfortunately, it does not seem to be reproducible. I have added a test here: https://github.com/near/rainbow-bridge/pull/190 It takes 2 blocks that were submitted successfully and 1 that failed to submit and resubmits it to the contract. I have even verified that the input to the contract is exactly the same by extracting the input to the transactions from etherscan. Unfortunately, it does not seem to reproduce the error: run tests from https://github.com/near/rainbow-bridge/pull/190 to reproduce.

We need to either:
* Call the current near-bridge contract deployed to Ropsten to get `last` field.
* Get the current state of the near-bridge contract deployed to Ropsten. See some hints on how it can be done: https://ethereum.stackexchange.com/questions/3353/is-there-an-easy-way-to-get-the-full-state-of-a-contract
And try replaying with it.
* Or replay all transactions that were submitted to this contract from the very beginning to recover its state.

Note. It might be possible the bug is that we are using `prev` instead of `last` here: https://github.com/near/rainbow-bridge/blob/4031aa81e9360f29b3ff05b6306d0a135469d238/libs-sol/nearbridge/contracts/NearBridge.sol#L162

@ailisp @bowenwang1996 WDYT?

ailisp:
Fix #191 and #169


environment/lib/near2eth-relay.js
```diff
@@ -104,18 +119,33 @@ class Near2EthRelay {
...
+                        continue;
+                    }
                     lightClientBlock = await this.near.connection.provider.sendJsonRpc('next_light_client_block', [lastFinalBlockHash]);
                     if (!lightClientBlock) {
                         await sleep(300);
+                        continue;
+                    }
+                    // Because fetch currentValidators and lightClientBlock isn't atomic, it's possible we happen to
+                    // fetch lightClentBlock cross epoch boundary. Fetch another time to ensure that's not the case.
+                    let currentValidatorsNow = await this.near.connection.sendJsonRpc('validators', [null]).result;
```
nearmax, line 137: We should not rely on three RPC requests `validators `, `next_light_client_block`, `validators` being sequentially consistent. Weird things, like implementation quirks of our RPC server or connecting to a different node can break it. Instead we should pass specific block height with  `validators`, if possible, AFAIK all our RPC endpoints allow that (except light client) CC @frol to confirm.

libs-sol/nearbridge/test/NearBridge2.js
```diff
@@ -128,3 +128,47 @@ contract('NearBridge2', function ([_, addr1]) {
 //     });
 // });
 
+contract('Test adding blocks in new epoch when bps change', function ([_, addr1]) {
```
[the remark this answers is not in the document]
ailisp: Yes I checked

environment/lib/near2eth-relay.js
ailisp: If don't use `validators`, we'd have to wait one epoch to know a block + the bp of this block here. Would be great we can know current epoch bps.

libs-sol/nearbridge/test/NearBridge2.js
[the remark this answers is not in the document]
ailisp: Check out to this https://github.com/near/rainbow-bridge/pull/192/commits/f5e7015b83a389e9f795bb4247b5a07b49c55496 to see test fails, with error message "number of BPs should match number of approvals"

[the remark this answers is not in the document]
nearmax: Sweet

environment/lib/near2eth-relay.js
bowenwang1996: yes you can pass a specific height here.

frol: 1. If you pass a block height instead of the `null`, you will get a view specifically pinned to the block height you request.
2. Use `.validators` method instead of a low-level `sendJsonRpc`.

ailisp: Thanks. Good to learn 1. `2` seems more straightforward here, will do that

ailisp:
Encountered a new kind of glitch in e2e: https://buildkite.com/nearprotocol/rainbow-bridge/builds/373#5eed625f-630c-4677-9d66-92bb5b32cf43
Error was createAccount rpc timeout, but when you retry, the previous createAccount become success, thus lead to "AccountAlreadyExist". I work around it here, but it probably considers to be a bug in near-api-js. A more reliable way would be use broadcast-txn-async to create account and check tx_status until success/fail. /cc @vgrichina

nearmax:
> Encountered a new kind of glitch in e2e: https://buildkite.com/nearprotocol/rainbow-bridge/builds/373#5eed625f-630c-4677-9d66-92bb5b32cf43
> Error was createAccount rpc timeout, but when you retry, the previous createAccount become success, thus lead to "AccountAlreadyExist". I work around it here, but it probably considers to be a bug in near-api-js. A more reliable way would be use broadcast-txn-async to create account and check tx_status until success/fail. /cc @vgrichina

@ailisp Please file an issue to near-api-js.

libs-sol/nearbridge/contracts/NearBridge.sol
```diff
@@ -212,7 +212,7 @@ contract NearBridge is INearBridge {
         for (uint i = 0; i < backup.next_bps_length; i++) {
             backup.next_bps[i] = last.next_bps[i];
         }
-        for (uint i = 0; i < backup.approvals_after_next.length; i++) {
```
ailisp: In solidity why we would also keep a var approvals_after_next_length? backup.approvals_after_next.length seems to work

nearmax: It does not compile.

ailisp: Got it

PR #202, opened 2020-07-20
fix 197

ailisp:
It now refund to txn signer instead of nearfuntoken


ailisp:
I tries switch account, use `--near-master-account node0`. but always getting:
```
TypedError: [-32000] Server error: Transaction 81HKpHvD62Ch6frJPba52RWP2F1Rv2ct5fQj4amgXbFq doesn't exist
```
txn submit (broadcast-tx-async) was successful. Why

PR #213, opened 2020-07-22
Split init method of Near2EthClient to reduce fees

nearmax:
Init method is too heavy and does not initialize on Ropsten


environment/index.js
```diff
@@ -95,6 +95,11 @@ RainbowConfig.declareOption('daemon', 'Whether the process should be launched as
 RainbowConfig.declareOption('bridge-src', 'Path to the rainbow-bridge source. Will use current repo if not provided.', path.join(__dirname, '../'));
 RainbowConfig.declareOption('core-src', 'Path to the nearcore source. It will be downloaded if not provided.', '');
 RainbowConfig.declareOption('nearup-src', 'Path to the nearup source. It will be downloaded if not provided.', '');
+RainbowConfig.declareOption(
+    'eth-gas-multiplier',
+    'How many times more in Ethereum gas are we willing to overpay.',
```
ailisp, line 100: this is clever!

environment/scripts/prepare.sh
```diff
@@ -20,9 +20,6 @@ then
 echo "near-core home not specified..."
 git clone "https://github.com/nearprotocol/nearcore" $CORE_SRC
 eval CURR_DIR=$(pwd)
-cd $CORE_SRC
-# Freeze nearcore version to avoid RPC changes breaking tests.
-git checkout df431768f95bc4117dc7c4d954e38f1b96fd3b9f
```
ailisp, line 25: we might consider bin search which commit cause near proof invalid in near prover

nearmax: Weird. It passes now. Maybe it was a fluke.

PR #220, opened 2020-07-23
Address 219

nearmax:


environment/lib/near2eth-relay.js
```diff
@@ -237,22 +237,10 @@ class Near2EthRelay {
                 }
                 try {
                     lightClientBlock = await near.connection.provider.sendJsonRpc('next_light_client_block', [clientBlockHash]);
-                    break;
-                } catch (err) {
```
ailisp, line 241: i suggest keep this catch, so we can tell it's near sendJsonRpc error or web3 error:
```
for ... {
    try {
            lightClientBlock = await near.connection.provider.sendJsonRpc('next_light_client_block', [clientBlockHash]);
    catch(err) {...}
    try{
        await clientContract.methods.addLightClientBlock(borshBlock)...
        break
    }catch(err) {...}
}
```

nearmax: The problem with this catch is that if for whatever reason we are getting wrong header and client rejects it we will not re-fetch the new header and will be trying to resubmit the broken header over and over.

ailisp: In the second catch put a `continue`?

nearmax:
I don't think it is needed anymore. Closing.

PR #227, opened 2020-07-27
Fix invalid proof error in ci e2e

ailisp:
Run on ci to get a invalid proof. happens frequently on ci but never locally


ailisp:
@nearmax fix turns out to be trivial: https://github.com/near/rainbow-bridge/pull/227/files#diff-14092c650910d5f422ac9c2324e10e88R64 😂
I can reproduce e2e on local ubuntu, and after the fix it pass. will undo other changes

nearmax:
> @nearmax fix turns out to be trivial: https://github.com/near/rainbow-bridge/pull/227/files#diff-14092c650910d5f422ac9c2324e10e88R64 😂
> I can reproduce e2e on local ubuntu, and after the fix it pass. will undo other changes

Nice! That's weird, I was not able to reproduce it locally when running e2e.sh, but it was failing almost consistently on CI. While both me locally and CI were using the most recent master. I would expect this kind of issue to cause either consistent pass or consistent failure.

ailisp:
> Nice! That's weird, I was not able to reproduce it locally when running e2e.sh, but it was failing almost consistently on CI. While both me locally and CI were using the most recent master. I would expect this kind of issue to cause either consistent pass or consistent failure.

Yeah the difference between mac and ubuntu is mac always give same timestamp and timestamp_nanosec

nearmax:
> > Nice! That's weird, I was not able to reproduce it locally when running e2e.sh, but it was failing almost consistently on CI. While both me locally and CI were using the most recent master. I would expect this kind of issue to cause either consistent pass or consistent failure.
> 
> Yeah the difference between mac and ubuntu is mac always give same timestamp and timestamp_nanosec

My mind is blown.

PR #230, opened 2020-07-29
Fix 221

### LINKED ISSUE #221: Memorize block hash and transaction hash when recording used events in TokenLocker
Currently TokenLocker memorizes the hash of the proof when it tries to prevent double spending. However there can be several valid proofs for the same event https://github.com/near/rainbow-bridge/blob/2fca92c314405b344b48359165dd262d2b910b72/libs-sol/token-locker/contracts/TokenLocker.sol#L55. We need to memorize  block hash and transaction hash instead.

ailisp:
fix #221

### DIFF
--- a/libs-sol/token-locker/contracts/TokenLocker.sol
+++ b/libs-sol/token-locker/contracts/TokenLocker.sol
@@ -17,7 +17,8 @@ contract TokenLocker {
     bytes public nearToken_;
     INearProver public prover_;
 
-    mapping(bytes32 => bool) public usedEvents_;
+    // blockHash -> OutcomeReciptId -> Used
+    mapping(bytes32 => mapping(bytes32 => bool)) public usedEvents_;
 
     event Locked(
         address indexed token,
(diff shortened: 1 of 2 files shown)


libs-sol/token-locker/contracts/TokenLocker.sol
```diff
@@ -50,16 +51,18 @@ contract TokenLocker {
...
 
         // Unpack the proof and extract the execution outcome.
         Borsh.Data memory borshData = Borsh.from(proofData);
         ProofDecoder.FullOutcomeProof memory fullOutcomeProof = borshData.decodeFullOutcomeProof();
         require(borshData.finished(), "Argument should be exact borsh serialization");
 
+        bytes32 blockHash = fullOutcomeProof.outcome_proof.block_hash;
+        bytes32 receiptId = fullOutcomeProof.outcome_proof.outcome_with_id.outcome.receipt_ids[0];
+        require(!usedEvents_[blockHash][receiptId], "The burn event cannot be reused");
+        usedEvents_[blockHash][receiptId] = true;
```
nearmax, line 64: AFAIR receipt ids are globally unique CC @evgenykuzyakov , so it is sufficient to memorize receipt id only.

ailisp: That's great!

> Also, please rebuild the contracts.

Token locker contracts is not commited: https://github.com/near/rainbow-bridge/tree/master/libs-sol/token-locker/dist

PR #231, opened 2020-07-29
remove dist files in nearbridge and nearprover

ailisp:
Just found token-locker ignore all dist files, which is helpful as solc compiles deterministically and we will not forget recompile contract (e2e will compile them). So doing same for nearbridge and nearprover


ailisp:
Close due to contracts move out

PR #234, opened 2020-07-30
Fix 208

### LINKED ISSUE #208: Burn half of the bond when challenge is successful
As reported by a third-party:

> Also, the callenge sends all the stake to the challenger, which means that the attacker can front-run and get all the money back. So to ensure that attacker is loosing money anyways, half of the stake can be burnt/sent to charity. And only half of the stake should go to the challenger.

ailisp:
Fix #208


nearmax:
@ailisp Let me know when this is closed.

Issue #238, opened 2020-08-03
e2e test for eth-relay catchup


nearmax:
We need to have a test that checks that the fast catchup of eth-relay works. We can achieve it by copying our existing e2e.sh test and placing a large delay before the eth-relay starts (additionally we might want to increase ganache block production speed to 1 block each 3 sec to exercise our eth-relay speed).

PR #254, opened 2020-08-04
Test ethrelay catchup and test challenge invalid block

ailisp:
Implement #238


environment/lib/near2eth-relay.js
```diff
@@ -170,6 +170,100 @@ class Near2EthRelay {
...
+                console.log(`Encountered error while requesting light client block ${err}`);
+                await sleep(1000);
+            }
+        }
+        console.log(`${JSON.stringify(lightClientBlock)}`);
+        while (true) {
+            let borshBlock = borshify(lightClientBlock);
+            console.log('Mutate block by one byte')
+            console.log(borshBlock)
+            borshBlock[Math.floor(borshBlock.length * Math.random())] += 1
```
nearmax, line 245: What if byte is already 255? Will it wrap around or saturate?

ci/test_challenge.sh
```diff
@@ -0,0 +1,65 @@
...
+sleep 5
+yarn run pm2 list
+node index.js start eth-relay
+sleep 5
+yarn run pm2 list
+node index.js start near-watchdog
+sleep 5
+yarn run pm2 list
+
+sleep 30
```
nearmax, line 50: Awesome, I see you had to increase the bond time to reduce flakiness.

environment/index.js
```diff
@@ -443,6 +444,24 @@ RainbowConfig.addOptions(
     ],
 );
 
+// Testing command
+const dangerCommand = program.command('DANGER');
```
nearmax, line 448: Could we add description to DANGER command and submit_invalid_near_block command?

environment/lib/near2eth-relay.js
ailisp: It wrap around:
```
> a=Buffer([255])
<Buffer ff>
> a[0]+=1
256
> a
<Buffer 00>
```

ci/test_challenge.sh
ailisp: Yeah, 10s lock time was too short to send the challenge

ailisp:
@nearmax got this error in e2e test chanllenge, this's also always happens locally (Seems there's either a problem with watchdog, or contract)
```
[2020-08-07T22:36:14Z] (node:5160) UnhandledPromiseRejectionWarning: Error: [-32000] Server error: 5GkUH4hzeZs17nrL8wSrsNLt7ZHJ8ApgXxrtRk6pD79v does not exist
...
[2020-08-07T22:36:14Z]     at async Command.execute (/var/lib/buildkite-agent/builds/buildkite-i-01dd6e880ebe52ccf-1/nearprotocol/rainbow-bridge/environment/commands/transfer-eth-erc20-from-near.js:148:24)
```
Underlying code is https://github.com/near/rainbow-bridge/blob/master/environment/commands/transfer-eth-erc20-from-near.js#L148

Do you have idea why a light_client_proof respond "does not exist" on requested receiptId? The receiptId is responded by rpc of the successful burn txn so it should exist. Also @bowenwang1996

nearmax:
> @nearmax got this error in e2e test chanllenge, this's also always happens locally (Seems there's either a problem with watchdog, or contract)
> 
> ```
> [2020-08-07T22:36:14Z] (node:5160) UnhandledPromiseRejectionWarning: Error: [-32000] Server error: 5GkUH4hzeZs17nrL8wSrsNLt7ZHJ8ApgXxrtRk6pD79v does not exist
> ...
> [2020-08-07T22:36:14Z]     at async Command.execute (/var/lib/buildkite-agent/builds/buildkite-i-01dd6e880ebe52ccf-1/nearprotocol/rainbow-bridge/environment/commands/transfer-eth-erc20-from-near.js:148:24)
> ```
> 
> Underlying code is https://github.com/near/rainbow-bridge/blob/master/environment/commands/transfer-eth-erc20-from-near.js#L148
> 
> Do you have idea why a light_client_proof respond "does not exist" on requested receiptId? The receiptId is responded by rpc of the successful burn txn so it should exist. Also @bowenwang1996

Is it 100% reproducible? Or does it happen randomly? There might be a race condition issue when ViewClient hasn't learnt yet that this receipt exists. I've seen similar issue with light client header -- RPC would tell me that this header exists, but if I try to immediately grab it with light client endpoint it would tell me it is not know, but then if I wait it becomes known.

bowenwang1996:
> Do you have idea why a light_client_proof respond "does not exist" on requested receiptId? The receiptId is responded by rpc of the successful burn txn so it should exist. Also @bowenwang1996

You need to wait for the next block for the outcome root.

ailisp:
> You need to wait for the next block for the outcome root.

The next block you mean "near block" right? I waited for a few minutes and retry the rpc manually and it still "does not exist".

ailisp:
> Is it 100% reproducible? Or does it happen randomly? There might be a race condition issue when ViewClient hasn't learnt yet that this receipt exists. I've seen similar issue with light client header -- RPC would tell me that this header exists, but if I try to immediately grab it with light client endpoint it would tell me it is not know, but then if I wait it becomes known.

Seems always reproducible locally, and wait doesn't help (tried print the params and retry rpc a few minutes later and just now (10+ hours later), still not exist.

bowenwang1996:
What does the set up look like? How many near nodes are there and what is the epoch length?

ailisp:
> What does the set up look like? How many near nodes are there and what is the epoch length?

It's default near init setup, one node, epoch length is 60

bowenwang1996:
I wonder whether it is because block production is too fast and garbage collection kicks in very quickly. Does it help if you change epoch length to 600 or make the node archival?

ailisp:
> I wonder whether it is because block production is too fast and garbage collection kicks in very quickly. Does it help if you change epoch length to 600 or make the node archival?

Good point!

nearmax:
> > I wonder whether it is because block production is too fast and garbage collection kicks in very quickly. Does it help if you change epoch length to 600 or make the node archival?
> 
> Good point!
> Update: it's probably the reason! passed first time, the restart with new config.json couldn't automated yet, but it passes. At the meantime, i've log and manually call rpc that verified every hash calculation is correct in NearBridge.sol. So we're good now. Thanks for help!

I've seen this issue before with bridge e2e tests, too bad it did not occur to me that this is the case. Thank you @bowenwang1996 !

PR #255, opened 2020-08-05
minor upds

Kouprin:


environment/commands/stop/all.js
```diff
@@ -0,0 +1,30 @@
...
+stop = function(name) {
+    try {
+        var a = A;
+        a._name = name;
+        StopManagedProcessCommand.execute(a);
+    } catch (err) {
+        console.log('Error stopping', name, err);
+    }
+}
+
```
Kouprin, line 16: Please suggest how to make it better.

environment/commands/stop/all.js
```diff
@@ -0,0 +1,30 @@
...
+    static execute() {
+        console.log('Stopping all processes...');
+
+        StopLocalNearNodeCommand.execute();
+        for (const name of ['ganache', 'eth-relay', 'near-relay', 'near-watchdog']) {
+            stop(name);
+        }
+
+        console.log('Stopping all processes done');
+    }
```
Kouprin, line 27: It doesn't exist at the end. Please suggest how to terminate node execution at the end.

index.js
```diff
@@ -0,0 +1 @@
+environment/index.js
```
Kouprin, line 1: I found this symlink is super useful.

environment/commands/stop/all.js
ailisp: For this particular case, you possibly need disconnect from ProcessManager.
In general, you can always force exit node by process.exit(0) for any kind of "not terminate" (we use this widely here). Or if you want a better way, call `process._getActiveRequests()` and `process._getActiveHandles()` to figure out anything block nodejs exit, and close them will unblock

index.js
ailisp: 👍

environment/commands/stop/all.js
ailisp: I think this is fine. For future housekeeping, refactor StopMangedProcessCommand to a function stop(name), and in here and StopManagedProcessCommand, call stop(name)

environment/commands/stop/all.js
ailisp:
```suggestion
        process.exit(0);
    }
```

Kouprin: @ailisp using process.exit(0) seems to be not correct as it terminates all subprocesses called in `StopManagedProcessCommand.execute(a);`. So we can't make sure that actual stopping has happened.

I'm personally okay to stop it by ctrl+c until we accept such bad code as creating empty class A. :)

ailisp: Ah i got the problem, so actual stop happens in a callback of StopManagedProcessCommand so we didn't know when it finishes. I suggest checking `pm2.list(errback)` in loop, before `console.log('Stopping all processes done');`, check result, sleep(1) if until everything stopped.

ailisp: Another way would be turn pm2 callback api to promise, so we are sure they stopped after `await pm2.stop(name)`:
https://nodejs.org/dist/latest-v12.x/docs/api/util.html#util_util_promisify_original
https://github.com/3axap4eHko/pm2-promise

Kouprin: @ailisp ok, I've tried several solutions you suggested. It seems I'm too unexperienced in nodejs as I cannot make anything of it working.
I've updated current PR with using pm2-promise as the most promisable solution :) with the code that personally makes sense to me. There is a traceback:
```
Error: process or namespace not found
    at /Users/kpr/node_modules/pm2/lib/API.js:1538:28
    at /Users/kpr/node_modules/pm2/lib/Client.js:734:12
    at /Users/kpr/node_modules/pm2-axon-rpc/lib/client.js:45:10
    at Parser.<anonymous> (/Users/kpr/node_modules/pm2-axon/lib/sockets/req.js:67:8)
    at Parser.emit (events.js:315:20)
    at Parser._write (/Users/kpr/node_modules/amp/lib/stream.js:91:16)
    at doWrite (_stream_writable.js:392:12)
    at writeOrBuffer (_stream_writable.js:374:5)
    at Parser.Writable.write (_stream_writable.js:314:12)
    at Socket.ondata (_stream_readable.js:708:22)
    at Socket.emit (events.js:315:20)
    at addChunk (_stream_readable.js:297:12)
    at readableAddChunk (_stream_readable.js:273:9)
    at Socket.Readable.push (_stream_readable.js:214:10)
    at Pipe.onStreamRead (internal/stream_base_commons.js:186:23)
```
Need help to finish the PR.

ailisp: This error `Error: process or namespace not found` happens when the process already stopped or didn't start before. You can safely try catch to ignore this case. I would do
```
try{
...
}catch(e){
console.log(e.message)
}
```
first to see what is the error message first, then modify the catch to be:
```
if (e.message.indexOf('process or namespace not found')>=0) {
    console.log('process xxx is stopped')
} // assume the message contains this, if not, adjust it.
```

Kouprin: It still makes no sense to me. Additional try/except around ProcessManager.delete(serviceName) didn't help. Traceback says that real executing of happens in guts of pm2 which in unreachable for try/catch.

ailisp: let me try your fork 🤔 we are able to catch this error in environment/process/commands/stop.js, but that's not pm2-promise

ailisp: I made it work now, problem is in pm2 promise it does an additional console.error(err), I removed it by fork it. And in your `stop` function, you need to return Promise ..., otherwise javascript returns null, and await stop() doesn't wait the promise finish. I change it to async/await from .then.catch, should be more alike to rust. 
➜  environment git:(upds) ✗ node index.js stop all             
Stopping all processes...
Stopping local near node...
ganache already stopped
eth-relay already stopped
near-relay already stopped
near-watchdog already stopped
Stopping all processes done
➜  environment git:(upds) ✗

Issue #257, opened 2020-08-06
Current system status


Kouprin:
We need a service that delivers the status of the system. @nearmax suggested to put it into index.js `status` command.
The service should return the following information:
1. Near node status
- Which network we use (local, testnet, mainnet)
- Is the node enabled and reachable
- Deployed and absent contracts
- Who is responsible for deploying those contracts
- Version of the node
2. Etherium node status
- Which network we use (ganache, ropsten, main)
- Is the node enabled and reachable
- Deployed and absent contracts
- Who is responsible for deploying those contracts
- Version of the node
3. Near relay status
- Is `near-relay` process running and its version
- How many NEAR are minted and burnt
4. Etherium relay status
- Is `eth-relay` process running and its version
- How many ETH are locked and unlocked
- How many TXs passed and rejected
5. Watchdog status
- Is `near-watchdog` process running and its version
- Number of challenges sent

PR #259, opened 2020-08-07
index.js status

Kouprin:
First version of `node index.js status` command. Related to #257.

Answer for turned ganache/near-node on:
Screen Shot 2020-08-09 at 2 58 46 PM

Answer for turned ganache/near-node off:
Screen Shot 2020-08-09 at 2 59 20 PM

Known issues:
1. Not checking bytecode of deployed ETH contracts. Not sure is it possible. Marked as TODO.
2. Not checking NEAR Contracts validity because of poor and unclear access to them. Marked as TODO.
3. No eth-relay/near-relay/watchdog status.

@nearmax, @ailisp if you can help with 1 or 2 please do.

### DIFF
--- a/environment/commands/status.js
+++ b/environment/commands/status.js
@@ -1,0 +1,415 @@
+const Web3 = require("web3");
+const nearlib = require("near-api-js");
+const fetch = require("node-fetch");
+const fs = require("fs");
+
+const { verifyAccountGently } = require("../lib/near-helpers");
+const { RainbowConfig } = require("../lib/config");
+const { normalizeEthKey } = require("../lib/robust");
+
+// Verdicts
+const Ok = "ok";
+const Info = "info";
+const Warn = "warn";
+const Error = "error";
+
+const Valid = "valid";
+const Deployed = "deployed";
+const Unknown = "unknown";
+const Unreachable = "unreachable";
+const Invalid = "invalid";
+const NotVerified = "not verified";
+const UsingMaster = "using data from master account";
+const RecordNotFound = "record not found";
+const ABINotFound = "abi not found";
+const ContractNotFound = "contract not found";
+const InternalError = "internal error";
+
+const NoColor = "\x1B[39m";
+
+// TODO please refactor to avoid async/await
+const request = async (url) => {
+  try {
+    const response = await fetch(url);
+    const json = await response.json();
+    return [
+      Ok,
+      JSON.stringify(json.version),
+      JSON.stringify(json.sync_info["latest_block_height"]),
+    ];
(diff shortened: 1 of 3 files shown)


Kouprin:
> Looks good so fat, would be helpful to also add near-relay/eth-relay/watchdog status from pm2

Absolutely. There is an issue https://github.com/near/rainbow-bridge/issues/257.

ailisp:
> Not checking NEAR Contracts validity because of poor and unclear access to them. Marked as TODO.

For near contract, they have initialized() method: https://github.com/near/rainbow-bridge/blob/master/libs-rs/eth-client/src/lib.rs#L151 try to call that method, and if contract is not present or not intialized, you'll get error.

> Not checking bytecode of deployed ETH contracts. Not sure is it possible. Marked as TODO.

Seems possible with this method:
https://web3js.readthedocs.io/en/v1.2.9/web3-eth.html#getcode

Kouprin:
> > Not checking NEAR Contracts validity because of poor and unclear access to them. Marked as TODO.
> 
> For near contract, they have initialized() method: https://github.com/near/rainbow-bridge/blob/master/libs-rs/eth-client/src/lib.rs#L151 try to call that method, and if contract is not present or not intialized, you'll get error.
> 
> > Not checking bytecode of deployed ETH contracts. Not sure is it possible. Marked as TODO.
> 
> Seems possible with this method:
> https://web3js.readthedocs.io/en/v1.2.9/web3-eth.html#getcode

Both solutions don't work for me.

First one is implemented specifically for Eth2NearClientContract. I don't know how to get access to it and do something like:
```
const contract = new nearlib.Contract(masterAccount, contractAccount, {});
const initialized = await contract.initialized();
```
As expected, I get an error:
```
TypeError: contract.initialized is not a function
```

The second one I use the following way:
```
console.log(await web3.eth.getCode(address));
```
and it always returns `0x0`.

Kouprin:
I'm merging this one and we'll figure out how to resolve minor issues later.

ailisp:
> const contract = new nearlib.Contract(masterAccount, contractAccount, {});
const initialized = await contract.initialized();

You need to list all method that contract have (or at least `init method`), example: https://github.com/near/rainbow-bridge/blob/master/environment/commands/init-near-fun-token.js#L39


> console.log(await web3.eth.getCode(address));

This sucks, seem i can get a lot of "web3 getCode get 0x0" search result but couldn't see a solution. Have you try ganache gui https://www.trufflesuite.com/ganache? They have a contract view, we may find out how they get the info from there source.

PR #263, opened 2020-08-11
init-eth-contracts + refactoring

Kouprin:
added .prettierrc + prettier record in package.json


ailisp:
Are there a prettier config file that can be included in this PR?

PR #264, opened 2020-08-11
status prints if near contract is deployed

Kouprin:
@ailisp Your ideas are very helpful. Finally I found the solution that do what's need to do. Is it possible to make it better? I'd prefer not to call fake function.


environment/commands/status.js
```diff
@@ -79,10 +79,25 @@ class NearContracts {
...
     }
     try {
-      const contract = new nearlib.Contract(masterAccount, contractAccount, {});
-      // TODO #257 check contract validity here
-      //console.log(contract);
-      return new Status(Unknown, Info, NotVerified);
+      const nearAccount = new nearlib.Account(near.connection, masterAccount);
+      const contract = new nearlib.Contract(nearAccount, contractAccount, {
+        changeMethods: ["boo"],
+        viewMethods: [],
```
ailisp, line 85: i think need "initialized" in viewMethods. Also i just found only eth-client contract defined this method, we need copy it to eth-prover and mintable-fungible-token contract

Kouprin:
I'm cherry-picking this to https://github.com/near/rainbow-bridge/pull/266. Let this way work until we update our contracts.

Issue #258, opened 2020-08-07
Comprehensive robustness enhancement


ailisp:
Here is a tracker of know robustness issue found in tests. Fix them as part of other PRs when they occur:
- [x] near-api-js query rpc timeout: https://buildkite.com/nearprotocol/nearcore/builds/2930#a5b21a48-d73f-40f3-ba4a-e3463f4c5110, fix in #254 
- [x] web3 nonce tracking error, unavoidable based on https://ethereum.stackexchange.com/questions/2527/is-there-a-way-to-find-an-accounts-current-transaction-nonce, fix by retry on higher nonce when it's failed due to nonce, in #254, 
- [ ] to fully avoid it need to replace all eth contract call with the wrapped RobustWeb3.callContract and replace all near-api-js query rpc to wrapped version, which isn't done yet.
- [x] near-watch dog didn't start: https://buildkite.com/nearprotocol/rainbow-bridge/builds/558#d42d8bde-0ecc-484b-a924-e9fd0acc4e87 practically it's not an issue but this is annoying on ci. Fixed in #273  
- [x] transfer-eth-erc20-from near uses a unwrapped version of nearcontract function call which didn't retry on timeout: https://buildkite.com/nearprotocol/rainbow-bridge/builds/586#f88f8696-8dd0-4a3c-8f48-244acc510261
Should be replaced with wrapped version (note it's not a borshContract as it's serialized in json)
- [ ] reach to near rpc failure not handled: https://buildkite.com/nearprotocol/rainbow-bridge/builds/672#dbf28908-6de8-438e-81e5-ab4ee36856b8

nearmax:
Thank you for creating this issue and owning the product.

ailisp:
@nearmax What was the way we used to ensure pm2 is started? It didn't happen for a while, but happens yesterday, and cause some e2e test fail,
this start near-watchdog didn't start without error (see pm2 list after it): https://buildkite.com/nearprotocol/rainbow-bridge/builds/560#e898c53a-ce75-4f30-9494-c6a3f41bfa24
this https://buildkite.com/nearprotocol/rainbow-bridge/builds/558#9bba6cf9-b6d7-47f4-aa24-c3116e697b63 ganache didn't start without error

djsatok:
> reach to near rpc failure not handled: https://buildkite.com/nearprotocol/rainbow-bridge/builds/672#dbf28908-6de8-438e-81e5-ab4ee36856b8
This is outdated

[...]

PR #267, opened 2020-08-13
docs: align Usage snippets; chmod +x; two robustness fixes; npm package test

chadoh:
* align Usage snippets: readme showed `npm i -g` option followed by local clone usage option `node index.js --help`. This makes `--help` option use global install version, then gives more info on local clone version.
* add `bin` section to `environment/package.json` so that using the global install will create an executable called `rainbow-bridge` (I think!)
* make `environment/index.js` executable, so you can execute with `./index.js` instead of `node index.js`. This required adding a `#!/usr/bin/env node` line to top of index.js.
* my editor is configured to automatically lint files according to project settings. Given that there is a `environment/.eslintrc` file, it used the settings specified in it. It seems like maybe this is not wanted?


ailisp:
@chadoh sorry i made a PR doing a subset of your PR, i'll take over this PR, resolve conflict and merge, sounds good?

> my editor is configured to automatically lint files according to project settings. Given that there is a environment/.eslintrc file, it used the settings specified in it. It seems like maybe this is not wanted?

@Kouprin proposed to use prettier, which is current consensus, so we would not use eslintrc. Other parts of your pr looks good to me.

ailisp:
Update: i added these things:
- add a packaging npm test, fix and test manually & automatically that npm install -g rainbow-bridge-cli work in all cases in e2e
- publish to npm
- fix two robustness issue, mentioned in #258 and #278

Kouprin (approved):
Awesome. There will be conflicts with https://github.com/near/rainbow-bridge/pull/266, please let me know if there is something important to merge that I can do improperly.

PR #279, opened 2020-08-19
validators ordered

Kouprin:


environment/lib/near2eth-relay.js
```diff
@@ -68,7 +68,7 @@ class Near2EthRelay {
         let currentValidators = null
         while (!lightClientBlock) {
           // @ts-ignore
-          currentValidators = await this.near.connection.provider.validators(
+          currentValidators = await this.near.connection.provider.validators_ordered(
```
nearmax, line 71: Gotta call RPC directly like this https://github.com/near/rainbow-bridge/blob/master/environment/lib/near2eth-relay.js#L363

nearmax:
FYI @ailisp

Issue #146, opened 2020-07-07
Mintable fungible token tests


nearmax:
Mintable fungible token extends regular fungible token and it reuses its tests. However, unlike regular fungible token it does not have fixed balance, this creates an entire set of corner cases that one need to explore with tests.

Also, we don't have test cross contract calls between mintable fungible token and prover.

ailisp:
done in rainbow-bridge-rs

PR #317, opened 2020-08-22
Add all mint related unit tests

ailisp:
Address #146, unit test all corner cases in mint and burn.


environment/commands/transfer-eth-erc20-to-near.js
```diff
@@ -195,6 +195,9 @@ class TransferETHERC20ToNear {
     )
     // @ts-ignore
     try {
+      console.log('=== proof locker')
```
nearmax, line 198: Should we remove this debug code?

libs-rs/mintable-fungible-token/src/lib.rs
```diff
@@ -590,6 +601,9 @@ mod tests {
     fn carol() -> AccountId {
         "carol.near".to_string()
     }
+    fn rainbow_bridge_eth_on_near_prover() -> AccountId {
```
nearmax, line 604: What is the purpose of this method?

libs-rs/mintable-fungible-token/src/lib.rs
```diff
@@ -27,7 +27,10 @@ Properties inherited from the standard Fungible Token:
 use borsh::{BorshDeserialize, BorshSerialize};
 use near_sdk::collections::{UnorderedMap, UnorderedSet};
 use near_sdk::json_types::U128;
-use near_sdk::{env, ext_contract, near_bindgen, AccountId, Balance, Promise, StorageUsage};
+use near_sdk::{
+    env, ext_contract, near_bindgen, AccountId, Balance, Promise, PromiseOrValue, StorageUsage,
+};
+use serde::Deserialize;
```
nearmax, line 33: What do we need serde for?

libs-rs/mintable-fungible-token/src/lib.rs
```diff
@@ -878,4 +892,171 @@ mod tests {
...
+            196, 199, 73, 127, 190, 26, 136, 104, 65, 161, 149, 165, 214, 34, 205, 96, 5, 60, 19,
+            118,
+        ];
+        context.storage_usage = env::storage_usage();
+
+        context.attached_deposit = 1000 * STORAGE_PRICE_PER_BYTE;
+        testing_env!(context.clone());
+
+        contract.burn(
+            1000.into(),
```
nearmax, line 912: Let's make it a local variable

libs-rs/mintable-fungible-token/src/lib.rs
```diff
@@ -878,4 +892,171 @@ mod tests {
...
+            "0123456789abcdef0123456789abcdef01234567".into(),
+        );
+        assert_eq!(contract.get_balance(alice()).0, total_supply - 1000);
+        assert_eq!(contract.get_total_supply().0, total_supply - 1000);
+    }
+
+    #[test]
+    #[should_panic(expected = "recipient should be a hex: OddLength")]
```
nearmax, line 920: Nice, thanks for this corner case test. Corner case tests are very important!

libs-rs/mintable-fungible-token/src/lib.rs
```diff
@@ -878,4 +892,171 @@ mod tests {
...
+        contract.locker_address = [
+            196, 199, 73, 127, 190, 26, 136, 104, 65, 161, 149, 165, 214, 34, 205, 96, 5, 60, 19,
+            118,
+        ];
+        context.storage_usage = env::storage_usage();
+
+        context.attached_deposit = 1000 * STORAGE_PRICE_PER_BYTE;
+        testing_env!(context.clone());
+
+        let proof: Proof = serde_json::from_reader(
```
nearmax, line 978: I see, why we need serde. We need to leave a comment that serde is only used in tests. We should also make sure that compiler successfully eliminates it during wasm compilation, by making sure that the contract size does not change noticeably after this PR.

libs-rs/mintable-fungible-token/src/lib.rs
```diff
@@ -878,4 +892,171 @@ mod tests {
...
+        contract.mint(proof);
+        assert_eq!(
+            contract.get_balance(rainbow_bridge_eth_on_near_prover()).0,
+            1000
+        );
+        assert_eq!(contract.get_total_supply().0, total_supply + 1000);
+    }
+
+    #[test]
+    #[should_panic(expected = "Event cannot be reused for minting.")]
```
nearmax, line 995: Nice, double-spending corner case.

environment/commands/transfer-eth-erc20-to-near.js
ailisp: Yes

libs-rs/mintable-fungible-token/src/lib.rs
ailisp: want to use this account, so follow the pattern of `alice()` `carol()` etc

libs-rs/mintable-fungible-token/src/lib.rs
ailisp: deserialize a proof dumped from eth

libs-rs/mintable-fungible-token/src/lib.rs
ailisp: Yep, I'm thinking of
```
#[not(cfg(test))]
#[derive(..., no SerdeDeserialize)]
struct Proof{
...
}

#[cfg(test)]
#[derive(..., SerdeDeserialize)]
struct Proof{
...
}
```

Do you know a better way to write this?

> We should also make sure that compiler successfully eliminates it during wasm compilation, by making sure that the contract size does not change noticeably after this PR.

Yeah if optimizer works well that's better

ailisp:
Contracts have been moved to rainbow-bridge-rs, so will address comment and recreate PR there

PR #333, opened 2020-09-01
Remove test flag

ailisp:
Fix https://github.com/near/rainbow-bridge-cli/issues/247
This has to be merge together with:
https://github.com/near/rainbow-bridge-lib/pull/5
https://github.com/near/rainbow-bridge-rs/pull/6

ci is expected to fail now, but after above two pr merged should ci must pass

### DIFF
--- a/commands/eth-dump.js
+++ b/commands/eth-dump.js
@@ -37,7 +37,7 @@ class ETHDump {
     }
     console.log(
       `Downloading block ${endBlock} down to ${startBlock} to ${path}. ${
-        endBlock - startBlock + 1
+      endBlock - startBlock + 1
       } blocks in total.`
     )
 
@@ -52,7 +52,7 @@ class ETHDump {
     try {
       // Only WebSocket provider can close.
       web3.currentProvider.connection.close()
-    } catch (e) {}
+    } catch (e) { }
     extractor.destroy()
   }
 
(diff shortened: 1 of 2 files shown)


Kouprin:
Where do we really validate ethash?

nearmax:
> > Where do we really validate ethash?
> 
> We always validate ethash and validate=False is not used even in tests, so i just drop it

We need this parameter for tests, because Ganache is not doing PoW, and therefore headers produced by Ganache do not have correct ethash. The same applies to PoA networks. When our e2e tests start Ganache, the start command writes this param  as false into config https://github.com/near/rainbow-bridge-cli/blob/master/commands/start/ganache.js#L31

nearmax (changes requested):
> > Where do we really validate ethash?
> 
> We always validate ethash and validate=False is not used even in tests, so i just drop it

We need this parameter for tests, because Ganache is not doing PoW, and therefore headers produced by Ganache do not have correct ethash. The same applies to PoA networks. When our e2e tests start Ganache, the start command writes this param  as false into config https://github.com/near/rainbow-bridge-cli/blob/master/commands/start/ganache.js#L31

So we should add compilation flag into https://github.com/near/rainbow-bridge-rs/pull/6 and move `near-client-validate-ethash` into `prepare` command.

Kouprin:
PR bankruptcy rule

PR #335, opened 2020-09-02
Adjustment to e2e tests so solidity contract works.

ailisp:
See manual ci run of this PR: https://buildkite.com/nearprotocol/rainbow-bridge/builds/725
it's "after merge https://github.com/near/rainbow-bridge-sol/pull/11" and run e2e tests.

Don't merge before https://github.com/near/rainbow-bridge-sol/pull/11 merge and new rainbow-bridge-sol is published

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,7 +222,7 @@ node index.js transfer-eth-erc20-from-near --amount 1 --near-sender-account rain
 
 You should observe the change of the ERC20 balance as reported by the CLI.
 
-## Contract Development Workflow
+## Contract and rainbow-bridge-lib Development Workflow
 
 Above steps are ways to run a local bridge and development workflows you need if make any changes to rainbow-bridge-cli. If you want to update any of solidity or rust contracts, they're not in this repo now and workflow is as following.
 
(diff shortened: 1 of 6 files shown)


Kouprin:
@ailisp if the test passing (as happening for now), it means you forget to increase rb-sol version. It will block the PR automatically and not allow to merge.

The invariant is the following: cross-repository updates should fail until their children have not introduced their new versions. If tests pass, you did it wrong.

ailisp:
@Kouprin you analysis is right on most of cases, but this PR's change doesn't introduce incompatible changes in e2e tests (The change works in both previous and current solidity contract)
What we must ensure is
- this should not merge before publish new raibnow-bridge-sol
- after publish new rainbow-bidge-sol and bump version here, CI must still pass

Kouprin (changes requested):
@ailisp the idea about versioning is deeper in this case. This case helped me to figure out better understanding.

The principle of versioning is: we must increase version following our logic, not because we add incompatible changes.

Technically, you may NOT increase version of rb-cli in this PR if it not breaks anything. However, we want to enforce yarn/npm to use new version of rb-sol. That's why it's important to increase the version of rb-cli and require only fresh rb-sol (which may not be introduced yet).

> What we must ensure is
> - this should not merge before publish new raibnow-bridge-sol
> - after publish new rainbow-bidge-sol and bump version here, CI must still pass

Version understanding, described above, helps to resolve both. You will not able to merge rb-cli while rb-sol is not introduced + you can restart CI after and make sure CI passes. BTW, there is no need to pass CI with previous version of rb-sol in general.

ailisp:
> Technically, you may NOT increase version of rb-cli in this PR if it not
breaks anything. However, we want to enforce yarn/npm to use new version of
rb-sol. That's why it's important to increase the version of rb-cli and
require only fresh rb-sol (which may not be introduced yet).
Right
>

   - this should not merge before publish new raibnow-bridge-sol
   - after publish new rainbow-bidge-sol and bump version here, CI must
   still pass

Yeah I think we have the same understanding on this problem. I documented
this here:
https://github.com/near/rainbow-bridge-cli/pull/335/files#diff-04c6e90faac2675aa89e2176d2eec7d8R245

> You will not able to merge rb-cli while rb-sol is not introduced + you
can restart CI after and make sure CI passes.
Unfortunately we didn't have way to enforce when in this case ci passed,
have to manually emphasis in PR title or put a "do not merge" label

Kouprin:
> Unfortunately we didn't have way to enforce when in this case ci passed,
> have to manually emphasis in PR title or put a "do not merge" label

@ailisp you don't need "do not merge" label. Please check https://github.com/near/rainbow-bridge-cli/pull/340

ailisp:
Ah i see, i didn't update version because new version was not available
until this is merged:
https://github.com/near/rainbow-bridge-sol/pull/11/files
And it's very clever to use the not existing version to make ci fail!

Kouprin:
waiting for merging https://github.com/near/rainbow-bridge-sol/pull/11

mfornet:
e2e tests are being executed in nayduck now

PR #342, opened 2020-09-10
use rainbow-token-connector

Kouprin:


package.json
```diff
@@ -30,11 +30,12 @@
     "node-fetch": "^2.6.0",
     "pm2": "^4.4.0",
     "pm2-promise": "^2.0.1",
-    "rainbow-bridge-lib": "1.0.2",
+    "rainbow-bridge-lib": "near/rainbow-bridge-lib#8e5567c0277d13430f952a19bb5e16ceb0b4392c",
```
ailisp, line 33: git version of dependency used to have issues in yarn, it incorrectly use a cached version to me unless i clear global yarn cache. Does it work for you?

ailisp (approved):

package.json
Kouprin: It works well on fresh machine or docker. In case of local running, ~/.cache folder must be deleted before `yarn install --force`.

PR #348, opened 2020-09-15
Add replace-duration parameter for NearClient

abacabadabacaba:


index.js
```diff
@@ -212,6 +212,11 @@ RainbowConfig.declareOption(
   'The challenge window during which anyone can challenge an incorrect ED25519 signature of the Near block, in EthClient, in seconds.',
   14400
 )
+RainbowConfig.declareOption(
```
nearmax, line 215: Please also add it to `init-near-contracts` list.

nearmax (changes requested):

nearmax (approved):
Actually approving, conditional on the comment being addressed

index.js
abacabadabacaba: You probably meant `init-eth-client`. Added.

ailisp (approved):

nearmax (approved):

PR #349, opened 2020-09-17
fix broken prepare in rainbow-bridge-cli npm package

ailisp:
it's only true when it's in a git repo. Otherwise this fail with exit code 128


frol (approved):
I confirm that this resolves the issue

PR #360, opened 2020-09-30
bump version to 2.0.0, deploy contract in rinkeby, tested locally and rinkeby

ailisp:


nearmax (changes requested):
This change is incorrect. We need to first publish 1.1.3 that reverts the token factory, and we need to do it for all rainbow-bridge repos. Then we need to publish 2.0.0 that introduces it.

ailisp:
transfer to near fully works. but from near has "proof is not valid error":
https://rinkeby.etherscan.io/tx/0x4c74014bc496232994612fcb49729749cfb283ab51f72cf3abf9967e60c9cfe6

still investigating

nearmax (approved):

nearmax (approved):
Verified that it works locally and on Rinkeby.

PR #362, opened 2020-09-30
Updating README.md to specify the location of near account secret key

mdzor:
Updating README.md to specify the location of near account secret key


Kouprin (commented):
This is valid only in case of TestNet.

mfornet (approved):

mfornet:
Closing since this is outdated, and we are on the process of re-writing README

PR #370, opened 2020-10-07
Pin nearup to use old v1 nearup release 0.1.2

chefsale:
Pin the nearup version to the 0.1.2 release which is nearup v1, so we can land nearup v2 onto master.  Create a release for nearup v1: https://github.com/near/nearup/releases/tag/0.1.2 and latest version of nearup v2 which should be installed via pip once @mhalambek finishes his PR, this is just for now.

### DIFF
--- a/scripts/prepare.sh
+++ b/scripts/prepare.sh
@@ -48,6 +48,7 @@ echo "Linking the specified local repo from ${LOCAL_NEARUP_SRC} to ${NEARUP_SRC}
 ln -s $LOCAL_NEARUP_SRC $NEARUP_SRC
 fi
 mkdir -p $NEARUP_LOGS
+cd $NEARUP_SRC && git checkout 0.1.2
 
 cd $CORE_SRC
 cargo build --package neard --bin neard


scripts/prepare.sh
```diff
@@ -47,6 +47,7 @@ else
 echo "Linking the specified local repo from ${LOCAL_NEARUP_SRC} to ${NEARUP_SRC}"
 ln -s $LOCAL_NEARUP_SRC $NEARUP_SRC
 fi
+cd $NEARUP_SRC && git checkout 0.1.2
```
frol, line 50: 1. It seems that by doing `cd` here you break the next commands
2. Why don't we just git clone the fixed tag in the first place?
3. It does not seem sane to force the local nearup_src to some tagged version instead of whatever development version it is there, does it?

frol (changes requested):

scripts/prepare.sh
chefsale: Yes, thats true will change that I was just about to do that :100:

chefsale: Ah, the problem is only that you have to checkout the commit on the LOCAL_NEARUP_SRC if somebody is using that so can't just clone. Reversed the order though now.

chefsale: The master after that commit is not compatible to new nearup interface so we have to force it, until @mhalambek creates a PR to download the pypi package and migrates the commands to the new interface instead of using a checkout of the repo.

frol (approved):

PR #402, opened 2020-10-27
extract sol contracts

Kouprin:


frol:
> extract sol contracts

@Kouprin Is the intent to include the contracts when the package is published to npm?

Kouprin:
> > extract sol contracts
> 
> @Kouprin Is the intent to include the contracts when the package is published to npm?

It includes only contract interfaces and makes the package smaller. I think it's what we want to have in this package.

frol:
@Kouprin Your words contradict the content of the PR :thinking: 

Do you want to exclude or include those *.sol files? This change *include* them into the package.

Kouprin:
> @Kouprin Your words contradict the content of the PR 🤔
> 
> Do you want to exclude or include those *.sol files? This change _include_ them into the package.

Sorry for my poor english. I want to extract only *.sol files into the package. No other files are needed.

frol:
I see it now. We want the `rainbow-bridge-sol` package to only contain those `*.sol` files. I have no experience with Solidity, so I would defer the decision on whether this is a useful package to someone with the relevant experience.

I confirm that the PR will lead to that the only files in the published package will be: `package.json`, `CHANGELOG.md`, and those `nearbridge/contracts/*.sol` and `nearprover/contracts/*.sol`.

frol (approved):

Kouprin:
> I have no experience with Solidity, so I would defer the decision on whether this is a useful package to someone with the relevant experience.

I checked how token-connector works. It uses only sol-files from the package, all other files seems to be useless to it.

frol:
I wonder if it only cares about interfaces (`INearBridge.sol`)

Otherwise, it looks good to me

Kouprin:
> I wonder if it only cares about interfaces (INearBridge.sol)

It uses also prover/INearProver.sol, prover/ProofDecoder.sol, bridge/NearDecoder.sol, bridge/Borsh.sol. We may revisit those dependencies later and make interfaces up.

PR #404, opened 2020-10-29
global package.json

Kouprin:


.gitmodules
```diff
@@ -1,4 +1,4 @@
-[submodule "eth2near/ethashproof"]
-	path = eth2near/ethashproof
+[submodule "ethashproof"]
+	path = ethashproof
```
frol, line 2: Should we move it back into `eth2near`? :smile:

DEV.md
```diff
@@ -0,0 +1 @@
+Using workspaces: https://classic.yarnpkg.com/en/docs/workspaces/
```
frol, line 1: Let's name this file `CONTRIBUTING.md` following the common practices.

write_config.js
```diff
@@ -0,0 +1,17 @@
+const { RainbowConfig } = require('rainbow-bridge-utils')
```
frol, line 1: This file is listed in .gitignore. Did you mean to still push it?

frol (approved):

.gitmodules
Kouprin: ++

DEV.md
Kouprin: ++

write_config.js
Kouprin: ++

cli/commands/status.js
```diff
@@ -339,8 +339,8 @@ class ServicesStatus {
   }
 
   processArgs (args) {
-    var res = []
-    for (var i = 0; i + 1 < args.length; i++) {
+    const res = []
+    for (let i = 0; i + 1 < args.length; i++) {
       if (args[i].startsWith('--')) {
         res.push(args[i].substring(2, args[i].length) + '=' + args[i + 1])
       }
```
Kouprin, line 346: is it okay to JS? const res; res.push(data)? It seems linter fixed this one automatically.

frol: It is valid JS... it is fine... `const` is only prevents the reassignment of the variable, but not the mutation of the object itself :laughing:

Kouprin: Not sure is it constant array or constant address of array.

frol (approved):

cli/commands/status.js
frol: It is constant address of array

Kouprin: Resolved.

PR #405, opened 2020-10-30
fix(dependencies): Local development wasn't working

mfornet:
After components refactorization into a single repository, a lot of dependencies and imports were broken.
This PR fixes all existing issues that were preventing local development. Notice that some of this issues were affecting production code too.

Test plan
=======

Test that run README step by step, and check that each one succeeded (Upgrade of previous `e2e.sh`)


utils/package.json
```diff
@@ -1,37 +1,37 @@
...
+        "lodash": "^4.17.20",
+        "near-api-js": "^0.26.0",
+        "web3": "=1.2.6"
+    },
+    "devDependencies": {
+        "standard": "*"
+    },
+    "scripts": {
+        "standard": "standard"
+    }
```
Kouprin, line 36: ?

Kouprin (approved):
You don't need to add yarn-error files.
yarn.lock also shouldn't be added because there is only yarn.lock in root directory.

utils/package.json
Kouprin: Can we also enable linter for json-files?

mfornet: I think we should do, for now I'll change my current json formatter to use 2 space indentation, since this seems to be what you are using.

PR #406, opened 2020-10-30
WIP - nayduck integration

Kouprin:


eth2near/eth2near-block-relay/index.js
```diff
@@ -168,8 +403,41 @@ class Eth2NearRelay {
...
+}
+
 exports.Eth2NearRelay = Eth2NearRelay
 exports.EthProofExtractor = EthProofExtractor
 exports.ethashproof = ethashproof
 exports.EthOnNearClientContract = EthOnNearClientContract
 exports.EthOnNearProverContract = EthOnNearProverContract
+exports.runEth2NearRelay = runEth2NearRelay
+
+require('make-runnable')
```
mfornet, line 443: Why require at the end of the file?

near2eth/near2eth-block-relay/index.js
```diff
@@ -14,13 +15,251 @@ const {
...
+
 class Near2EthRelay {
   async initialize () {
     // @ts-ignore
     this.robustWeb3 = new RobustWeb3(RainbowConfig.getParam('eth-node-url'))
     this.web3 = this.robustWeb3.web3
     this.ethMasterAccount = this.web3.eth.accounts.privateKeyToAccount(
-      normalizeEthKey(RainbowConfig.getParam('eth-master-sk'))
+      // TODO remove this hard-coded constant!
+      normalizeEthKey('0x2bdd21761a483f71054e14f5b827213567971c676928d9a1808cbfa4b7501201')
```
mfornet, line 261: Why do we need hardcoded constant instead of receiving it as argument?

eth2near/eth2near-block-relay/index.js
Kouprin: It's how `make-runnable` works. :)
It can only make functions runnable that are exported above.

mfornet (approved):

near2eth/near2eth-block-relay/index.js
Kouprin: Yep, this is bad. We should use config there when it will be available. For now I will pass it as argument.

Kouprin: Ok, I moved it into `runNear2EthRelay` temporary. This should be fine for now. It will be changed when new config will come.

PR #417, opened 2020-11-12
Upgrade nearup to latest version

mhalambek:
Instead of pulling nearup from git
use latest version from pip.

### DIFF
--- a/cli/commands/clean.js
+++ b/cli/commands/clean.js
@@ -1,8 +1,5 @@
 const { execSync } = require('child_process')
 const ProcessManager = require('pm2')
-const { existsSync } = require('fs')
-const { homedir } = require('os')
-const path = require('path')
 
 class CleanCommand {
   static execute () {
@@ -18,13 +15,11 @@ class CleanCommand {
           process.exit(1)
         }
         ProcessManager.disconnect()
-        if (existsSync(path.join(homedir(), '.rainbow', 'nearup', 'main.py'))) {
-          try {
-            console.log('Stopping nearup')
-            execSync('python3 ~/.rainbow/nearup/main.py stop')
-          } catch (err) {
-            console.log(`Error stopping nearup ${err}`)
-          }
+        try {
+          console.log('Stopping nearup')
+          execSync('$HOME/.local/bin/nearup stop')
+        } catch (err) {
+          console.error(`Error stopping nearup ${err}`)
         }
         console.log('Cleaning ~/.rainbow and ~/.near/localnet directory...')
         execSync('rm -rf ~/.rainbow ~/.near/localnet')
--- a/cli/commands/start/near.js
+++ b/cli/commands/start/near.js
@@ -7,7 +7,7 @@ const { RainbowConfig } = require('rainbow-bridge-utils')
 class StartLocalNearNodeCommand {
   static execute () {
     const command = util.format(
-      'python3 ~/.rainbow/nearup/main.py localnet --num-nodes 1 --binary-path %s',
+      '$HOME/.local/bin/nearup run localnet --num-nodes 1 --binary-path %s',
       '~/.rainbow/core/target/debug'
     )
     request(getLocalNearNodeURL(), { json: true }, (err, _res, _body) => {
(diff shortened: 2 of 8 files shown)


mhalambek:
Still left installing nearup in ci environment

chefsale:
Why do we need to install nearup in ci env or what do you exavtly mean by that?

mhalambek:
> Why do we need to install nearup in ci env or what do you exavtly mean by that?

Yes, nearup has to be installed for test

cli/commands/clean.js
```diff
@@ -18,13 +15,11 @@ class CleanCommand {
...
-        if (existsSync(path.join(homedir(), '.rainbow', 'nearup', 'main.py'))) {
-          try {
-            console.log('Stopping nearup')
-            execSync('python3 ~/.rainbow/nearup/main.py stop')
-          } catch (err) {
-            console.log(`Error stopping nearup ${err}`)
-          }
+        try {
+          console.log('Stopping nearup')
+          execSync('nearup stop')
```
chefsale, line 20: The only problem I see here it's not that you add this to the ci:
```
export PATH=$HOME/.local/bin:$PATH
```
the same problem will happen to normal bridge users, as if the don't have this in the path, the nearup command will fail as it is not in the in the path, I think you will have to use the full path for that reason.

cli/commands/start/near.js
```diff
@@ -7,7 +7,7 @@ const { RainbowConfig } = require('rainbow-bridge-utils')
 class StartLocalNearNodeCommand {
   static execute () {
     const command = util.format(
-      'python3 ~/.rainbow/nearup/main.py localnet --num-nodes 1 --binary-path %s',
+      'nearup run localnet --num-nodes 1 --binary-path %s',
```
chefsale, line 10: same as above

cli/commands/stop/process.js
```diff
@@ -3,12 +3,12 @@ const { execSync } = require('child_process')
 
 async function stopLocalNearNode () {
   console.log('Stopping near node')
-  const command = 'python3 ~/.rainbow/nearup/main.py stop'
+  const command = 'nearup stop'
```
chefsale, line 6: same as above

testing/ci/e2e_ci_prepare_env.sh
```diff
@@ -7,6 +7,9 @@ if [[ ! -d ~/go ]]; then
     wget -q -O - https://raw.githubusercontent.com/canha/golang-tools-install-script/master/goinstall.sh | bash
 fi
 
+pip3 install nearup --user
```
chefsale, line 10: this shouldn't be done in the ci, this is already done in bridge

testing/ci/e2e_ci_prepare_env.sh
```diff
@@ -7,6 +7,9 @@ if [[ ! -d ~/go ]]; then
...
+
 export GOROOT=~/.go
 export GOPATH=~/go
 export PATH=$GOPATH/bin:$GOROOT/bin:$PATH
+export PATH=$HOME/.local/bin:$PATH
```
chefsale, line 15: this should be added here, as the problem will persist for normal users, explained above

chefsale (changes requested):

cli/commands/clean.js
mhalambek: Regular bridge users can install it through pip3 as usual added note about in the README
They still require install of Rust to run anything, don't see as big issue installing nearup through pip then

For regular users docker should be used, which is unfortunately not working at the moment

mhalambek: Done, other way around, waiting for test to pass, looks ok now

chefsale (approved):
Yes, this looks good. Thanks for fixing it. :)

cli/commands/clean.js
mfornet: @chefsale @mhalambek this is not working for me locally since nearup is not being installed in this location `$HOME/.local/bin`. If we are installing `nearup` from pip I think it is safe to assume that nearup will be in the path.

chefsale: Of you run the prepare it should be installed in that folder. Have you initialize it with that?

PR #425, opened 2020-11-19
Fix 424

### LINKED ISSUE #424: Potentially wrong logs by near2eth-relay
While running bridge on Mainnet I have observed the following logs from near2eth-relay:
```
2020-11-18 19:59:45.349: Checking whether client is initialized.
2020-11-18 19:59:45.752: Client is initialized.
2020-11-18 19:59:46.532: Trying to submit new block at height 22780235.
2020-11-18 20:04:40.650: Submitted.
2020-11-18 20:04:41.152: Client height is 22737035, chain height is 22780235. Sleeping for 600 seconds.
2020-11-18 20:14:41.670: Client height is 22737035, chain height is 22780235. Sleeping for 600 seconds.
2020-11-18 20:24:42.152: Client height is 22737035, chain height is 22780235. Sleeping for 600 seconds.
```
The first line was printed out right after near2eth-relay was restarted. Two things look weird:
1. Relay reports successful submission of header height `22780235` but then it reports that client is at height `22737035`, even though it should've changed to `22780235`;
2. Three times it reports that the chain height is `22780235`, even though it should've changed since NEAR is producing 1 block a second.

nearmax:
Fix #424

### DIFF
--- a/near2eth/near2eth-block-relay/index.js
+++ b/near2eth/near2eth-block-relay/index.js
@@ -274,7 +274,7 @@ class Near2EthRelay {
         }
         delay = Math.max(delay, minDelay)
         console.log(
-          `Client height is ${bridgeState.currentHeight}, chain height is ${lastBlock.inner_lite.height}. Sleeping for ${delay} seconds.`
+          `Last valid header on the client: ${bridgeState.currentHeight}. Waiting for block ${lastBlock.inner_lite.height} to become valid for ${delay} seconds.`
         )
         clientHeightGauge.set(Number(BigInt(bridgeState.currentHeight)))
         chainHeightGauge.set(Number(BigInt(lastBlock.inner_lite.height)))
--- a/near2eth/near2eth-block-relay/index.js
+++ b/near2eth/near2eth-block-relay/index.js
@@ -273,11 +273,12 @@ class Near2EthRelay {
           )
         }
         delay = Math.max(delay, minDelay)
-        console.log(
-          `Last valid header on the client: ${bridgeState.currentHeight}. Waiting for block ${lastBlock.inner_lite.height} to become valid for ${delay} seconds.`
-        )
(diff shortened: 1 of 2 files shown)


near2eth/near2eth-block-relay/index.js
```diff
@@ -267,7 +267,7 @@ class Near2EthRelay {
         }
         delay = Math.max(delay, minDelay)
         console.log(
-          `Client height is ${bridgeState.currentHeight}, chain height is ${lastBlock.inner_lite.height}. Sleeping for ${delay} seconds.`
+          `Last valid header on the client: ${bridgeState.currentHeight}. Waiting for block ${lastBlock.inner_lite.height} to become valid for ${delay} seconds.`
```
abacabadabacaba, line 270: This message is not accurate either. `lastBlock.inner_lite.height` isn't necessarily the same as the height of the current unverified block. Rather, it's the height of the block that could have been submitted if there were no unverified block. These values are usually the same during catchup (both correspond to the last block in the epoch that follows the epoch of the current verified block), but not during normal operation.

near2eth/near2eth-block-relay/index.js
```diff
@@ -266,9 +266,11 @@ class Near2EthRelay {
...
         }
         delay = Math.max(delay, minDelay)
-        console.log(
-          `Client height is ${bridgeState.currentHeight}, chain height is ${lastBlock.inner_lite.height}. Sleeping for ${delay} seconds.`
-        )
+
+        const status = await this.near.connection.provider.sendJsonRpc('status', '')
+
+        console.log(`Last valid header on the client: ${bridgeState.currentHeight}. Next light client block: ${lastBlock.inner_lite.height}`)
+        console.log(`Chain height: ${status.sync_info.latest_block_height} Sleeping for ${delay} seconds.`)
```
mfornet, line 273: @abacabadabacaba can you check new messages make sense

PR #434, opened 2020-12-02
feat(adapter): allow minting test ERC20 tokens from CLI and adapter

Kouprin:


testing/adapter/index.js
```diff
@@ -24,11 +39,12 @@ function getEthErc20Balance ({ ethAccountAddress, ethNodeUrl, ethErc20Address, e
     JSON.parse(fs.readFileSync(ethErc20AbiPath)),
     remove0x(ethErc20Address)
   )
-  return ethContract.methods.balanceOf(remove0x(ethAccountAddress)).call().then(result => {
+  ethContract.methods.balanceOf(remove0x(ethAccountAddress)).call().then(result => {
```
frol, line 42: Why did you remove `return`? This will effectively mean that the async calls will get executed in background and you won't be able to await on the result from this function. Why do we need to explicitly call `exit()` here?

SkidanovAlex (approved):

testing/adapter/index.js
```diff
@@ -7,6 +7,21 @@ const {
...
+    JSON.parse(fs.readFileSync(ethErc20AbiPath)),
+    remove0x(ethErc20Address)
+  )
+  ethContract.methods
+    .mint(ethAccountAddress, Number(amount))
+    .send({ from: ethAccountAddress, gas: 5000000 }).then(result => {
+      console.log('OK')
+      exit(0)
+    })
+}
```
frol, line 23: FYI, this is how we should be able to rewrite this function in async style (it does not matter in this case):

```suggestion
async function mintErc20 ({ ethAccountAddress, amount, ethNodeUrl, ethErc20Address, ethErc20AbiPath }) {
  const robustWeb3 = new RobustWeb3(ethNodeUrl)
  const web3 = robustWeb3.web3
  const ethContract = new web3.eth.Contract(
    JSON.parse(fs.readFileSync(ethErc20AbiPath)),
    remove0x(ethErc20Address)
  )
  const result = await ethContract.methods
    .mint(ethAccountAddress, Number(amount))
    .send({ from: ethAccountAddress, gas: 5000000 });
  console.log('OK')
  exit(0)
}
```

frol (approved):
LGTM

Kouprin:
@frol the way of using these methods is the following:
- call the method by CLI from python lib
- wait until method finished
- get its stdout as the result of execution

The `exit(0)` is important because executing any of this CLI commands from command line leads to getting stuck forever. The other important thing is the process should finish after getting promise result to print to stdout the outcome, so it cannot be async and terminates before its children are finished.

If you know how to change this behavior to the better please assign an issue on me.

testing/adapter/index.js
Kouprin: Described in separated comment above.
`return` is not needed because the result of the function is void.

frol:
@Kouprin I am suspecting that the reason the node application "stuck" is due to the open websocket (web3.js) connection. We have used explicit `.close()` in other places and it helped.

PR #436, opened 2020-12-07
feat(cli, adapter): update CLI and adapter

Kouprin:
- revisit adapter methods
- add adapter methods for approve and lock eth-to-near txs
- expose token name in adapter methods
- add eth-to-near-find-proof method to CLI
- move transfer commands to testing CLI explicitly

### DIFF
--- a/eth2near/eth2near-block-relay/eth-proof-extractor.js
+++ b/eth2near/eth2near-block-relay/eth-proof-extractor.js
@@ -1,8 +1,9 @@
 const Tree = require('merkle-patricia-tree')
 const { Header, Proof, Receipt, Log } = require('eth-object')
+const utils = require('ethereumjs-util')
 const { encode } = require('eth-util-lite')
 const { promisfy } = require('promisfy')
-const { RobustWeb3 } = require('rainbow-bridge-utils')
+const { RobustWeb3, JSONreplacer } = require('rainbow-bridge-utils')
 
 function receiptFromWeb3 (result) {
   return Receipt.fromWeb3(result)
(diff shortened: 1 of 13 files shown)


utils/index.js
```diff
@@ -28,10 +28,18 @@ function getScript (name) {
   return path.resolve(path.join(__dirname, `scripts/${name}.sh`))
 }
 
+function JSONreplacer (key, value) {
+  if (typeof value === 'object' && value !== null && value.type === 'Buffer') {
+    return value.data
+  }
+  return value
+}
```
frol, line 36: Let's have a better name for it. I was really puzzled reading this function name, unused `key` argument, and the cryptic logic.

eth2near/eth2near-block-relay/eth-proof-extractor.js
```diff
@@ -67,6 +68,65 @@ class EthProofExtractor {
...
+        log = receiptLog
+        break
+      }
+    }
+    if (logFound) {
+      const logEntryData = logFromWeb3(log).serialize()
+      const receiptIndex = proof.txIndex
+      const receiptData = receiptFromWeb3(receipt).serialize()
+      const headerData = proof.header_rlp
+      const _proof = []
```
frol, line 108: Can it have a better name?

frol (approved):

utils/index.js
Kouprin: Let me add a comment there.

eth2near/eth2near-block-relay/eth-proof-extractor.js
Kouprin: This is taken from `from-near.js`. We can rename it.

Kouprin: Done.

utils/index.js
Kouprin: Added.

Issue #418, opened 2020-11-12
[Bug] Watchdog should ignore skipped signatures


nearmax:
Since we altered the behavior of NearOnEthClient when it verifies the signatures: https://github.com/near/rainbow-bridge/blob/master/contracts/eth/nearbridge/contracts/NearBridge.sol#L120 it is not breaking the watchdog that expects `checkBlockProducerSignatureInHead` to never fail as long as it stays within the range of the signatures: https://github.com/near/rainbow-bridge/blob/master/near2eth/watchdog/index.js#L68

abacabadabacaba:
What's the bug? Currently, `checkBlockProducerSignatureInHead` will revert if the signature is skipped, and the check `!e.reason || !e.signature` is specifically to ignore reverts (but not other exceptions).

nearmax:
@abacabadabacaba , it seems it was the intention, but when I run watchdog as is it crashes when it reaches skipped signature. Something does not work as intended.

SkidanovAlex:
> it seems it was the intention, but when I run watchdog as is it crashes when it reaches skipped signature. Something does not work as intended.

Can you provide more information about the failure? The stack trace?

mfornet:
Solved on #454

PR #454, opened 2021-01-17
fix(watchdog): Don't fail on skipped signature

mfornet:
When some of the blocks are not signed by all validators, bridge watchdog is unable to check the full block. Skip missing signatures.


mfornet:
Fixed #418

Issue #459, opened 2021-01-22
Check contracts are compiled deterministically


mfornet:
Add following test to CI:

Compile all contracts deterministically and check that output matches exactly the contracts in the repository.

sept-en:
Assign this task to me, please. I will take a look.

mfornet:
- Done for Ethereum contracts here: https://github.com/near/rainbow-bridge/blob/fix_ci/.buildkite/contracts-eth.yml#L2
- Done for NEAR contracts here: https://github.com/near/rainbow-bridge/blob/121d2629aa615763a75e15ebb8eeaeaf696108fd/contracts/near/.buildkite/pipeline.yml#L46

mfornet:
Done ethereum side on: https://github.com/near/rainbow-bridge/blob/master/.buildkite/contracts-eth.yml#L2

PR #489, opened 2021-02-09
fix: CI

mfornet:
Run all implemented tests on CI (tests from other repositories after merging into monorepo were not being executed)

Test plan
=======
- All tests that were running before merging into monorepo should be executed
- All tests are passing (CI is green)


abacabadabacaba:
I think that it is better to submit unrelated changes as separate pull requests.

mfornet:
> I think that it is better to submit unrelated changes as separate pull requests.

@abacabadabacaba Sure, the main purpose of this PR was fixing the CI. And just started by fixing some typos to trigger current CI. Still I see that I made several changes unrelated to fixing CI, so will split them in several PR

mfornet:
Fixes #459

PR #495, opened 2021-02-12
refactor: Use linter for solidity code

mfornet:
Test plan
=========
- linter test should pass


abacabadabacaba (changes requested):
I think that auto-formatting is a good idea, but I don't like the code that is produced with the default settings. Please add a prettierrc file with the following options:
```json
"printWidth": 100,
"explicitTypes": "preserve"
```
Also, please exclude generated files (*.full.sol).

abacabadabacaba (approved):
LGTM, although we may consider raising the line length to 120 to have even less stuff split into multiple lines.

PR #496, opened 2021-02-12
Auto upgrade bridge

mfornet:
Both ethereum contracts (nearbridge and nearprover) have auto-upgradability functionalities that allows single entity (admin) to call arbitrary function on it. This aligns with our current upgradability mode at this time, and will be improved in the future as we move toward a more decentralized setup.

Test Plan
=======
- CI should pass


mfornet:
Note created Draft PR for now, since #489 should be merged first.

contracts/eth/nearbridge/contracts/NearBridge.sol
```diff
@@ -79,27 +80,36 @@ contract NearBridge is INearBridge {
         Ed25519 ed,
         uint256 lockEthAmount_,
         uint256 lockDuration_,
-        uint256 replaceDuration_
+        uint256 replaceDuration_,
+        address admin_
     ) public {
+        require(replaceDuration_ > lockDuration_.mul(1000000000));
```
abacabadabacaba, line 86: Currently, because of this line, the unit tests fail with the following error:

```
Error: while migrating NearBridge: The contract code couldn't be stored, please check your gas limit.
```

(despite what the message says, it is revert, not out of gas). Please fix this.

abacabadabacaba (changes requested):

contracts/eth/nearbridge/contracts/NearBridge.sol
mfornet: Fixed

contracts/eth/nearbridge/contracts/AdminControlled.sol
```diff
@@ -0,0 +1,38 @@
+pragma solidity ^0.6;
+
+contract AdminControlled {
+    address public admin;
+    uint public paused;
+
```
sept-en, line 6: Add the constructor there, please. Thus the approach of setting admin in inherited classes will be smoother and we can also set the initial flags (which functions are paused on the initialization) in the ctor.

djsatok: Please add to the constructor the ability to set flags

PR #510, opened 2021-03-02
Add option to pause Near contracts

mfornet:
Test plan
=======
- (TODO): Near contracts are compiled deterministically
- (TODO): Unit test that stress pause conditions
- e2e pass

### DIFF
--- a/contracts/near/admin-controlled/src/lib.rs
+++ b/contracts/near/admin-controlled/src/lib.rs
@@ -1,0 +1,30 @@
+use near_sdk::env;
+
+pub type Mask = u128;
+
+pub trait AdminControlled {
+    fn is_owner(&self) -> bool {
+        env::current_account_id() == env::signer_account_id()
+    }
+
+    fn assert_owner(&self) {
+        assert!(self.is_owner());
+    }
+
+    /// Return the current mask representing all paused events.
+    fn get_paused(&self) -> Mask;
+
+    /// Update mask with all paused events.
+    /// Implementor is responsible for guaranteeing that this function can only be
+    /// called by owner of the contract.
+    fn set_paused(&mut self, paused: Mask);
+
+    /// Return if the contract is paused for the current flag and user
+    fn is_paused(&self, flag: Mask) -> bool {
+        (self.get_paused() & flag) != 0 && !self.is_owner()
+    }
+
+    fn check_not_paused(&self, flag: Mask) {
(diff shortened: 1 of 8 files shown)


contracts/near/eth-client/src/lib.rs
```diff
@@ -58,6 +59,8 @@ pub struct HeaderInfo {
     pub number: u64,
 }
```
sept-en, line 61: I suggest is that we either add another flag that makes it clear how to unpause everything (i.e. `const UNPAUSE_ALL: u128 = 0;` as we do similar on the Solidity side) or add one more method to the trait which - `setAllUnpaused()`.

contracts/near/admin-controlled/src/lib.rs
```diff
@@ -0,0 +1,30 @@
+use near_sdk::env;
+
+pub type Mask = u128;
```
sept-en, line 3: Is there any reason to use `u128` for the mask? I think even if we extend it one day to manage any functionality in contracts (not only plausibility) is still too huge. IMO the `u8` is enough for this, though we can extend it to `u16` or `u32` (for future functionality).

contracts/near/eth-client/src/lib.rs
```diff
@@ -58,6 +59,8 @@ pub struct HeaderInfo {
...
 
+const PAUSE_ADD_BLOCK_HEADER: u128 = 1;
```
djsatok, line 62: `u128` -> `Mask`

contracts/near/eth-prover/src/lib.rs
```diff
@@ -45,6 +47,8 @@ impl Default for EthProver {
     }
 }
 
+const PAUSE_VERIFY: u128 = 1;
```
djsatok, line 50: `u128` -> `Mask`

djsatok (changes requested):

djsatok:
Closes #509

contracts/near/eth-client/src/lib.rs
mfornet: This code should probably live on the service interacting with the contract, but not in the contract itself give that it is not used.

contracts/near/admin-controlled/src/lib.rs
mfornet: It is better to be on the safe side. Notice this is generic interface that can be implemented for any contract.

PR #518, opened 2021-03-06
Add adminSstoreWithMask to AdminControlled.sol

abacabadabacaba:


contracts/eth/nearbridge/contracts/AdminControlled.sol
```diff
@@ -31,6 +29,13 @@ contract AdminControlled {
         }
     }
 
+    function adminSstoreWithMask(uint key, uint value, uint mask) public onlyAdmin {
```
mfornet, line 32: Can you add a comment about how this works? and how it should be used

mfornet (approved):

PR #524, opened 2021-03-11
fix: Don't use env::signer_account_id

mfornet:
Instead use env::predecessor_account_id. Using signer_account_id
has the same problems of using tx.origin in Ethereum. See link
below for more details:

https://github.com/ethereum/solidity/issues/683

### DIFF
--- a/contracts/near/admin-controlled/src/macros.rs
+++ b/contracts/near/admin-controlled/src/macros.rs
@@ -1,0 +1,21 @@
+#[macro_export]
+macro_rules! impl_admin_controlled {
+    ($contract: ident, $paused: ident) => {
+        use admin_controlled::{AdminControlled as AdminControlledInner, Mask as MaskInner};
+        use near_sdk as near_sdk_inner;
+
+        #[near_bindgen]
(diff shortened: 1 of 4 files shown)


contracts/near/admin-controlled/src/macros.rs
```diff
@@ -0,0 +1,21 @@
+#[macro_export]
+macro_rules! impl_admin_controlled {
+    ($contract: ident, $paused: ident) => {
+        use admin_controlled::{AdminControlled as AdminControlledInner, Mask as MaskInner};
+        use near_sdk as near_sdk_inner;
```
abacabadabacaba, line 5: Using `use` in macros is not a very good idea, because it silently imports names into the surrounding namespace. Also, it assumes that `near_sdk` and `admin_controlled` crates are present in the list of dependencies of the current crate under their original names.

A more robust solution would be to not use `use` and to use fully qualifying names starting with `$crate` (because `$crate` is the only top-level name such that you know for sure what it refers to).

abacabadabacaba (changes requested):

PR #525, opened 2021-03-11
 fix: Use fixed near-sdk-version

mfornet:
Need to use fixed near-sdk-version for compatibility
with rainbow-token-connector. This restriction will
be lifted once the special commit used by the connector
lands on near-sdk-rs.

Note: merge after #524


abacabadabacaba:
Is this change needed to allow token connector to use `admin_controlled`?

mfornet:
> Is this change needed to allow token connector to use `admin_controlled`?

Yes

PR #532, opened 2021-03-16
fix: Use proper ethash argument from CLI

mfornet:
nearClientValidateEthash was mispelled, and as a consequence, the argument was always passed as false


cli/init/near-contracts.js
```diff
@@ -101,7 +101,7 @@ class InitNearContracts {
       hashesGcThreshold,
       finalizedGcThreshold,
       numConfirmations,
-      nearClientValidateEthhash === 'true',
+      nearClientValidateEthash === 'true',
```
abacabadabacaba, line 104: We should add code to make it fail if the value of `nearClientValidateEthash` is neither `'true'` nor `'false'`, to prevent errors (such as misspellings).

abacabadabacaba (changes requested):

PR #553, opened 2021-04-09
Various optimizations and fixes in Ethereum contracts

abacabadabacaba:
* NearBridge gas usage is reduced.
* Removed support for ECDSA signatures, which are not used.
* Compiler warnings fixed.
* Initialization functions are only callable by the admin.
* Some refactoring.


contracts/eth/nearbridge/contracts/NearBridge.sol
```diff
@@ -12,57 +13,39 @@ contract NearBridge is INearBridge, AdminControlled {
...
-
-    // Block producers of the current epoch.
-    BlockProducerInfo currentBlockProducers;
-    // Block producers of the next epoch.
-    BlockProducerInfo nextBlockProducers;
+    Ed25519 immutable edwards;
 
-    // The most recent head that is guaranteed to be valid.
-    BlockInfo head;
+    Epoch[3] epochs;
```
mfornet, line 35: Why using size `3` instead of `2`? 

In case there is a good reason I'm missing, what about using size `4`, so we avoid using `% 3` through the code, and instead use `& 3` (which is the same as `% 4).

contracts/eth/nearbridge/contracts/NearBridge.sol
```diff
@@ -202,161 +150,140 @@ contract NearBridge is INearBridge, AdminControlled {
...
+        for ((uint i, uint cnt) = (0, thisEpoch.numBPs); i != cnt; ++i) {
+            bytes32 stakes = thisEpoch.packedStakes[i >> 1];
+            if (nearBlock.approvals_after_next[i].some) {
+                votedFor += uint128(bytes16(stakes));
+            }
+            if (++i == cnt) {
+                break;
+            }
+            if (nearBlock.approvals_after_next[i].some) {
+                votedFor += uint128(uint256(stakes));
```
mfornet, line 215: I'm trying to figure out still if this is correct. My thought why it might be incorrect:

stakes is 32 bytes: [firstStake (16 bytes), secondStake (16 bytes)]

it is read as a 256bit integer **big-endian** (is it read as big endian?) so first stake is stored in the least significant bits.

After casting to 128, we are left again with `firstStake` rather than with `secondStake`.

mfornet (commented):
Haven't finished reviewing this. I have studied and learned a lot about solidity doing this review.

mfornet (approved):
Finish review with help from @abacabadabacaba 
LGTM, thanks for this optimization

contracts/eth/nearbridge/contracts/NearBridge.sol
mfornet: The current implementation is correct.

PR #556, opened 2021-04-14
Fix bug verifying merkle-patricia-tree proofs from ethereum

mfornet:
Our code had several unhandled edge cases while verifying the proofs, and some valid proofs weren't accepted. This PR reimplements the verification in a simpler and more robust way.

Sources used for reference:
- [Ethereum Yellow Paper](https://ethereum.github.io/yellowpaper/paper.pdf)
- [Ethereum MPT Wiki Page](https://eth.wiki/en/fundamentals/patricia-tree)

Added proof of concept [in python](https://gist.github.com/mfornet/0ff283274c0162f1cca45966bccf69ee) with similar implementation:


contracts/near/eth-prover/src/lib.rs
```diff
@@ -162,160 +149,155 @@ impl EthProver {
...
+            assert_eq!(path.as_slice(), &key[key_index..key_index + path.len()]);
+
+            if head >= 2 {
+                // Extension node
+                assert_eq!(proof_index + 1, proof.len());
+                assert_eq!(key_index + path.len(), key.len());
+                get_vec(&node, 1)
             } else {
-                panic!("This should not be reached if the proof has the correct format");
+                // Leaf node
```
abacabadabacaba, line 237: Your comments are wrong (you need to swap `Extension node` and `Leaf node`).

contracts/near/eth-prover/src/lib.rs
```diff
@@ -4,9 +4,6 @@ use eth_types::*;
 use near_sdk::{env, ext_contract, near_bindgen, Gas, PanicOnDefault, PromiseOrValue};
 use rlp::Rlp;
 
-#[cfg(test)]
-mod tests;
-
```
abacabadabacaba, line 9: Now there is a file `tests.rs` which is unused.

abacabadabacaba (approved):
LGTM (but see the comments below).

contracts/near/eth-prover/src/lib.rs
mfornet: Ohh, I see, thanks for pointing that. Will move the test over there.

Issue #555, opened 2021-04-12
Get rid of truffle-flattener


abacabadabacaba:
truffle-flattener is not maintained, and we have various kludges to make it work. We should get rid of it for good.

Issue #541, opened 2021-03-17
Build assets in CI instead of adding them to the repository


mfornet:
This issue applies equally to [rainbow-token-connector repository](https://github.com/near/rainbow-token-connector/).

The deployment experience is affected by the fact that we require binaries in the repository (from solidity and rust contracts) to match binaries compiled in a reproducible way. CI makes sure this is the case, to avoid using outdated binaries. 

The way it works has some drawbacks.

1. All binaries are added to the history of this repository, which is not ideal.
2. Developers need to recompile binaries if the made any change to some of the contracts.

Step 2 is particularly problematic since binaries should be compiled in a reproducible way. To achieve that we are using docker and start compilation from scratch. One of the problem with this, is that we can't reuse any cache, and compilation from scratch with docker is usually much slower than compiling natively in the developer machine.

To this end I propose removing compiled binaries from the repository and instead generating them in the CI and have them accessible through some endpoints ideally directly in github. In addition we should have a tool to download this assets depending on the current commit / branch / release version. This way developers don't need to compile contracts each time, and assets are still accessible on deployment time.

PR #564, opened 2021-04-18
Remove compiled Ethereum contracts from version control and switch to using artifact files

### LINKED ISSUE #491: Warnings compiling solidity contracts
Compilation of solidity contracts emits several warnings. To see this locally use:

```
cd contracts/eth/nearbridge
./dist.sh
```

### Warnings

- [x] [truffle-flattener](https://github.com/nomiclabs/truffle-flattener) is using old version of solidity-parser. This warning is already solved: https://github.com/solidity-parser/parser/issues/19

```
 Warning: Accessing non-existent property 'INVALID_ALT_NUMBER' of module exports inside circular dependency
```

- [ ] [No License](https://forum.openzeppelin.com/t/solidity-0-6-8-introduces-spdx-license-identifiers/2859) on solidity code triggers warning. Consider using [this approach](https://github.com/nomiclabs/truffle-flattener/issues/55#issuecomment-679157423) (Only keep one License rather than removing all of them).
```
dist/Borsh.full.sol: Warning: SPDX license identifier not provided in source file. Before publishing, consider adding a comment containing "SPDX-License-Identifier: <SPDX-License>" to each source file. Use "SPDX-License-Identifier: UNLICENSED" for non-open-source code. Please see https://spdx.org for more information.
```

- [x] Some docstring [will stop working](https://blog.soliditylang.org/2020/07/28/solidity-v0.7.0-release-announcement/) on version 0.7.0 of solidity
```
dist/NearBridge.full.sol:529:17: Warning: Only state variables can have a docstring. This will be disallowed in 0.7.0.
        bytes32 epoch_id;           /// Epoch start hash of this block's epoch. Used for retrieving validator information
                ^------^
```

- [x] `Warning: Function state mutability can be restricted to pure`

```
dist/NearDecoder.full.sol:464:5: Warning: Function state mutability can be restricted to pure
    function decodeInitialValidators(Borsh.Data memory data) internal view returns(InitialValidators memory validators) {
    ^ (Relevant source part starts here and spans across multiple lines).
```

- [ ] Installing gyp/scrypt [view log](https://gist.github.com/mfornet/dec42dd22418396cd4c31dd74b61bf6e)

abacabadabacaba:
Fixes #491, #555.
Partial fix for #541.


cli/init/eth-contracts.js
```diff
@@ -27,12 +28,17 @@ class EthContractInitializer {
       ethMasterAccount = ethMasterAccount.address
 
       console.log('Deploying ETH contract')
-      const tokenContract = new web3.eth.Contract(
-        JSON.parse(fs.readFileSync(ethContractAbiPath))
-      )
+      let abi, bytecode
+      if (ethContractArtifactPath) {
+        ({ abi, bytecode } = JSON.parse(fs.readFileSync(ethContractArtifactPath)))
```
mfornet, line 33: Are externals parenthesis required here?

mfornet (approved):
TIL that solidity compilers _guarantees_ that compilation is deterministic on every platform.

https://ethereum.stackexchange.com/a/9887/45323

mfornet:
One problem I see with current approach is that it is easier to mess up the right contract to use during deployments and interaction. Because it is possible to have different version on the contract and on the repository at the same time. For properly solving #541 we should publish instead these contracts in this repository github release page, and download it from there.

cli/init/eth-contracts.js
abacabadabacaba: Yes, see [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#assignment_without_declaration).

abacabadabacaba:
I think there is a more general problem. Suppose that we change the API for relayers, and modify the relayers' code accordingly. Now, if someone is using the new relayer code to interact with the old contract (or vise versa), it will break with some confusing error messages. So, just publishing the contracts is not the full solution, we need to add some check to make sure that the contracts the relayers are interacting with are compatible with the relayers' version.

PR #580, opened 2021-05-11
Use binary search instead of backtracking to find common root

mfornet:
The binary search function will try to look in the neighbourhood of the estimated value, rather than starting from the endpoints on the interval. This noticeably improves the speed of convergence (to O(1) most of the time), since after the first step binary search we can estimate the following values very precisely.


eth2near/eth2near-block-relay/index.js
```diff
@@ -40,8 +40,45 @@ function ethashproof (command, _callback) {
...
+  while (lo + 1 < hi) {
+    let mid = Math.floor((lo + hi) / 2);
+    if (await predicate(mid)) {
+      hi = mid;
+    } else {
+      lo = mid;
+    }
+  }
+  return hi;
+}
```
sept-en, line 78: I'm not sure if this is essential if we forbid Ropsten. This is most likely not required for Goerli and Mainnet.
Anyway, this is a good optimization, so it's anyway really great to have it!

sept-en (approved):
LGTM!

sept-en:
NB: this should be merged only after the chain of the following PRs is merged: #600, #602, #615

Issue #588, opened 2021-05-22
Cannot compile solidity project due to symlink (Hardhat)


vince0656:
Due to this issue discovered:

https://github.com/nomiclabs/hardhat/issues/1474

and the symlink found here:

https://github.com/aurora-is-near/rainbow-bridge/tree/master/contracts/eth/nearprover/contracts

I can't compile a project that uses the latest solidity 0.8^ rainbow bridge contracts

PR #589, opened 2021-05-23
Hardhat: remove symlink usage for deps. Fix HH408.

sept-en:
Currently, we have an issue with repositories that use the `rainbow-bridge` as the dependency, getting the HH306 error.
The real reason for that is that Hardhat is not able to correctly work with files defined outside of the project and not imported via module. In the commit [898800](https://github.com/aurora-is-near/rainbow-bridge/commit/898800c3c91b03105b858dd88c4f55e8d1450666) the fix for error `HH408` was introduced, trying to fix the limitation with the symlink. Unfortunately, exactly this leads to this HH306 error for the repositories, who use the `rainbow-bridge`, because the Hardhat has issues working with symlinks as well, as reported in #588.

The solution for the issue is to use the package instead of the relative path in Solidity contracts. Currently, I added the package inside the `package.json` that points to the appropriate module using the relative path. However, the current solution requires repositories who use it, apply the `postinstall` patch that changes the `nearbridge` module to the full path of the `rainbow-bridge` repo. The better solution would be to either release `nearbridge` inner module separately, so the `nearprover` will be able to point to this module, or try using the `gitpkg`, maybe this will be helpful as well.

Until this is finalized, which won't require everyone who use this repo to apply patches and having that much effort, this will be a draft PR.

### DIFF
--- a/.buildkite/contracts-eth.yml
+++ b/.buildkite/contracts-eth.yml
@@ -12,6 +12,7 @@ steps:
       command: |
           source testing/ci/e2e_ci_prepare_env.sh
           cd contracts/eth/nearprover
+          yarn --cwd ../
           yarn
           yarn test
       agents:
--- a/.github/workflows/contracts-eth.yml
+++ b/.github/workflows/contracts-eth.yml
@@ -21,6 +21,7 @@ jobs:
           - job-name: nearprover test
             cmd: |
               cd contracts/eth/nearprover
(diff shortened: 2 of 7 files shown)


mfornet:
@sept-en what is the status of the PR. Should I review?

mfornet (approved):

karim-en (approved):

PR #598, opened 2021-06-23
EthClient: fix tests.

### LINKED ISSUE #595: Error the trait `Serialize` is not implemented for `DoubleNodeWithMerkleProof`
Hello!
I got an error when I try to run the eth-client tests using `./test.sh`

qqqq

sept-en:
Fixes #595

### DIFF
--- a/contracts/near/eth-client/Cargo.toml
+++ b/contracts/near/eth-client/Cargo.toml
@@ -21,10 +21,13 @@ ethash = { git = "https://github.com/nearprotocol/rust-ethash" }
 hex = "0.4.0"
 rustc-hex = "2.1.0"
 
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+# serde's Serialize and Deserialize traits are required for `near_bindgen` macro for non-wasm32 targets
+serde = { version = "1.0", features = ["derive"] }
+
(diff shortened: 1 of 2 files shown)


contracts/near/eth-client/Cargo.toml
```diff
@@ -21,10 +21,13 @@ ethash = { git = "https://github.com/nearprotocol/rust-ethash" }
 hex = "0.4.0"
 rustc-hex = "2.1.0"
 
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+# serde's Serialize and Deserialize traits are required for `near_bindgen` macro for non-wasm32 targets
+serde = { version = "1.0", features = ["derive"] }
+
```
mfornet, line 27: TIL You can have features flags in Cargo.toml

mfornet (approved):
Good hack!

contracts/near/eth-client/Cargo.toml
sept-en: @mfornet I didn't check if we can have feature flags there, but I expect this will work only for targets
https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#build-dependencies

PR #600, opened 2021-06-29
Add support for EIP-1559

mfornet:


eth2near/ethashproof/ethash/consensus.go
```diff
@@ -570,14 +572,14 @@ func (ethash *Ethash) Finalize(chain consensus.ChainReader, header *types.Header
 	header.Root = state.IntermediateRoot(chain.Config().IsEIP158(header.Number))
 
 	// Header seems complete, assemble into a block and return
-	return types.NewBlock(header, txs, uncles, receipts), nil
+	return types.NewBlock(header, txs, uncles, receipts, trie.NewStackTrie(nil)), nil
```
sept-en, line 575: why do we add `trie.NewStackTrie(nil)` here?

sept-en (approved):
LGTM!

eth2near/ethashproof/ethash/consensus.go
mfornet: We are using a newer version of go-ethereum and this is a new required argument.

sept-en (approved):
LGTM!

NB: this should be merged only after #602 and #615 are merged.

PR #603, opened 2021-07-01
Use patched version from ethereumjs block

mfornet:
while https://github.com/ethereumjs/ethereumjs-monorepo/pull/1330 is merged


mfornet:
This doesn't work as expected.

PR #604, opened 2021-07-01
Rainbow on Binance Smart Chain

3alpha:
# New BSC Feature

The new BSC feature allows Rainbow bridge to run on Binance Smart Chain.

# Technical

When the Bridge is configured to use **bsc** mode, the first header has to be an epoch header (block_number%200==0). That is because the epoch header contains the validator set inside the `extra_data` field.  After that, bridge starts to receive and validate the other blocks. If current block at the time of starting the bridge isn't the epoch block, bridge gets initialised at previous epoch block.
Every time Near client gets a new epoch header it saves it and uses the new validator set from it.

# New Configurations:

The Bridge handles POW(ethash), POSA(bsc), or POA, by default the POW is used.

To use `bsc`:
```json
    nearClientValidateHeader: 'true',
    nearClientValidateHeaderMode: 'bsc'
```

and to use `ethash`:
```json
    nearClientValidateHeader: 'true',
    nearClientValidateHeaderMode: 'ethash'
```

here is an example of the new config file `~/.rainbow/config.json`:
```json
    {
        "nearNodeUrl": "http://localhost:3030",
        "nearNetworkId": "local",
        "nearMasterAccount": "node0",
        "nearMasterSk": "ed25519:3D4YudUQRE39Lc4JHghuB5WM8kbgDDa34mnrEP5DdTApVH81af7e2dWgNPEaiQfdJnZq1CNPp5im4Rg5b733oiMP",
        "ethNodeUrl": "https://data-seed-prebsc-1-s1.binance.org:8545",
        "ethMasterSk": "0x2bdd21761a483f71054e14f5b827213567971c676928d9a1808cbfa4b7501200",
        "nearClientValidateHeader": "true", 
        "nearClientValidateHeaderMode": "bsc",
        "nearClientSk": "ed25519:3D4YudUQRE39Lc4JHghuB5WM8kbgDDa34mnrEP5DdTApVH81af7e2dWgNPEaiQfdJnZq1CNPp5im4Rg5b733oiMP",
        "nearProverSk": "ed25519:3D4YudUQRE39Lc4JHghuB5WM8kbgDDa34mnrEP5DdTApVH81af7e2dWgNPEaiQfdJnZq1CNPp5im4Rg5b733oiMP",
        "ethEd25519Address": "0x62ff99964d1026bfd3fea2086b9d141917f1a793",
        "ethClientLockEthAmount": "1000",
        "ethClientLockDuration": "10",
        "ethClientAddress": "0x2cdaaa435937ae65705b97fbd01d28f4f6494545",
        "ethProverAddress": "0x5cdc96ce31bdf08ef39d5f65a9fd5e236ce2245f",
        "ethErc20Address": "0x7ef88a3901494435700ccc4c7615903b43ec2dbc",
        "ethLockerAddress": "0x24d718f7d8af7d7a4abaa2b1cb7abbb36a23a00d",
        "ethAdminAddress": "0xDf08F82De32B8d460adbE8D72043E3a7e25A3B39"
    }
```

# Deploy the bridge locally with BSC Mode

To run the bridge locally with Binance Smart Chain, run the following cmd: 

1. init yarn and generate Ethereum contracts: 

```bash
make init
```

2. Start Near blockchain and connect with Binance testnet.
```bash
make start-bsc
```

3. Deploy Near and Ethereum contracts

```bash
make full-bsc-contracts
```
or use already deployed contracts (on BSC), first copy past the following parameters inside the `~/.rainbow/config.json` file: 

```bash
    "ethEd25519Address": "0x62ff99964d1026bfd3fea2086b9d141917f1a793",
    "ethClientLockEthAmount": "1000",
    "ethClientLockDuration": "10",
    "ethClientAddress": "0x2cdaaa435937ae65705b97fbd01d28f4f6494545",
    "ethProverAddress": "0x5cdc96ce31bdf08ef39d5f65a9fd5e236ce2245f",
    "ethErc20Address": "0x7ef88a3901494435700ccc4c7615903b43ec2dbc",
    "ethLockerAddress": "0x24d718f7d8af7d7a4abaa2b1cb7abbb36a23a00d",
    "ethAdminAddress": "0xDf08F82De32B8d460adbE8D72043E3a7e25A3B39"
```
then run the following command:

```bash
make light-bsc-contracts
```

4. Start the relayers:

```bash
make start-relayer
```

5. When you are done you can stop the bridge by running:

```bash
make stop-all
```

- The bridge uses pm2 tool so you can use it to check logs and list the processes


cli/commands/start/binance-smart-chain.js
```diff
@@ -0,0 +1,17 @@
...
+    }
+
+    return {
+      ethNodeUrl: ethNodeUrl !== '' ? ethNodeUrl : 'https://data-seed-prebsc-1-s1.binance.org:8545',
+      ethMasterSk: ethMasterSk !== '' ? ethMasterSk : '0x2bdd21761a483f71054e14f5b827213567971c676928d9a1808cbfa4b7501200',
+      nearClientValidateHeader: nearClientValidateHeader,
+      nearClientValidateHeaderMode: 'bsc'
+    }
+  }
+}
```
mfornet, line 15: Is this missing something? Or the idea is that this command only sets up the configuration?

cli/commands/start/binance-smart-chain.js
```diff
@@ -0,0 +1,17 @@
+// Binance Smart chain configurations to connect with testnet
+class StartBinanceSmartChainNodeCommand {
+  static async execute ({ ethNodeUrl, ethMasterSk, nearClientValidateHeader }) {
+    if (nearClientValidateHeader !== 'true' && nearClientValidateHeader !== 'false') {
+      nearClientValidateHeader = 'true'
+    }
+
+    return {
+      ethNodeUrl: ethNodeUrl !== '' ? ethNodeUrl : 'https://data-seed-prebsc-1-s1.binance.org:8545',
+      ethMasterSk: ethMasterSk !== '' ? ethMasterSk : '0x2bdd21761a483f71054e14f5b827213567971c676928d9a1808cbfa4b7501200',
```
mfornet, line 10: Where is this secret key coming from? is this a good default? should we leave this empty and force the user to set the secret key.

cli/commands/start/binance-smart-chain.js
```diff
@@ -0,0 +1,17 @@
+// Binance Smart chain configurations to connect with testnet
+class StartBinanceSmartChainNodeCommand {
+  static async execute ({ ethNodeUrl, ethMasterSk, nearClientValidateHeader }) {
+    if (nearClientValidateHeader !== 'true' && nearClientValidateHeader !== 'false') {
+      nearClientValidateHeader = 'true'
+    }
+
+    return {
+      ethNodeUrl: ethNodeUrl !== '' ? ethNodeUrl : 'https://data-seed-prebsc-1-s1.binance.org:8545',
```
mfornet, line 9: Is this endpoint for public use? otherwise probably we want to remove it from here, and add in the docs how to get the binance endpoint.

contracts/near/eth-client/src/lib.rs
```diff
@@ -105,27 +109,44 @@ pub struct EthClient {
     trusted_signer: Option<AccountId>,
     /// Mask determining all paused functions
     paused: Mask,
+    /// Store the bsc epoch header key 
+    epoch_header: H256,
+    /// chain id
+    chain_id: u64,
```
mfornet, line 115: Since the state of the contract is changing we can't redeploy on top of the current contracts as is.

We need to migrate the state first! Some resources:
- https://stackoverflow.com/a/68191607/4950797
- https://github.com/near/NEPs/pull/123

We either need to do this atomically or pause the contract before hand, to avoid inconsistencies.

mfornet: This is needed only because we don't want to maintain two different versions of the client, though we are having two different versions deployed one for Ethereum and one for BSC.

contracts/near/eth-client/src/lib.rs
```diff
@@ -348,12 +374,189 @@ impl EthClient {
...
         dag_nodes: &[DoubleNodeWithMerkleProof],
+    ) -> bool {
+        if self.validate_header_mode == String::from("ethash") {
+            // validate ethash POW headers
+            return self.verify_header_pow(&header, &prev, &dag_nodes);
+        }else if self.validate_header_mode == String::from("bsc") {
+            // validate bsc POSA headers 
+            return self.verify_header_bsc(&header, &prev);
+        }
+        false
```
mfornet, line 390:
```suggestion
        match self.validate_header_mode {
            ...
        }
```

contracts/near/eth-client/src/lib.rs
```diff
@@ -348,12 +374,189 @@ impl EthClient {
...
+            
+            if !(is_turn == header.author) && header.difficulty != U256(diff_no_turn.into()){
+                return false;
+            }
+        }
+        true
+    }
+
+    /// Verify PoW of the header.
+    fn verify_header_pow(
```
mfornet, line 555:
```suggestion
    fn verify_header_ethash(
```

contracts/near/eth-client/src/lib.rs
```diff
@@ -421,6 +620,23 @@ impl EthClient {
         );
 
         (H256(pair.0), H256(pair.1))
+    }  
+
+    // check if the author is in the validators set.
+    fn is_in_validator_set (&self, epoch_header: &BlockHeader, add: Address) -> bool{
```
mfornet, line 626: The rust code doesn't look properly formatted. Can you use rustfmt to format it!

contracts/near/eth-client/src/tests.rs
```diff
@@ -525,13 +599,15 @@ fn add_two_blocks_from_400000() {
...
         true,
+        String::from("ethash"),
         400_000 / 30000,
         vec![blocks_with_proofs.first().unwrap().merkle_root],
         blocks[0].clone(),
         30,
         10,
         10,
         None,
+        97,
```
mfornet, line 610: Let's use named ChainId here, to be clear what this number actually means. In this case since we are using `ethash` let's use chain_id 3, though it is not used anywhere so it makes no real difference.

Just as a note for other reviewing this PR I found a list of chain id from networks supporting EVM: [Chainlist](https://chainlist.org/).

Makefile
```diff
@@ -0,0 +1,66 @@
...
+stop-all:
+	cli/index.js stop all
+
+build-eth-client:
+	cd contracts/near/eth-client && sudo ./build.sh
+
+test-eth-client:
+	cd contracts/near/eth-client && ./test.sh
+
+.PHONY: help init yarn-init gen-contracts start-bsc full-bsc-contracts light-bsc-contracts start-relayer stop-all build-eth-client test-eth-client start-ethash
```
mfornet, line 66: Thanks for the Makefile, it is a really helpful addition. 🎉

mfornet (changes requested):
For now only reviewed the code affecting the previous setup, will continue reviewing the code introducing the new feature next. I'm making this checkpoint to leave some comments.

cli/commands/start/binance-smart-chain.js
idirall22: its only for BSC configuration sets up.

cli/commands/start/binance-smart-chain.js
idirall22: Yes, it's a BSC testnet endpoint.

cli/commands/start/binance-smart-chain.js
idirall22: The SK is the same SK used for testing with ganache. This SK is used in many places not only for bsc but also for ethhash.

contracts/near/eth-client/src/lib.rs
```diff
@@ -348,12 +417,190 @@ impl EthClient {
...
+        address.copy_from_slice(&result[12..]);
+
+        H160(address.into()) == header.author
+    }
+
+    fn get_epoch_header(&self) -> BlockHeader {
+        self.headers.get(&self.epoch_header).unwrap()
+    }
+
+    // check if the auther address is valid and is in the validator set.
```
mfornet, line 558:
```suggestion
    // check if the author address is valid and is in the validator set.
```

contracts/near/eth-client/src/lib.rs
```diff
@@ -348,12 +417,190 @@ impl EthClient {
...
+        }
+
+        // skip difficulty verification
+        if self.validate_header {
+            // Ensure that the difficulty corresponds to the turn-ness of the signer
+            let validators = epoch_header.extra_data
+                [extra_vanity..(epoch_header.extra_data.len() - extra_seal)]
+                .to_vec();
+
+            // Get validator offset postion.
```
mfornet, line 579:
```suggestion
            // Get validator offset position.
```

contracts/near/eth-client/src/lib.rs
```diff
@@ -348,12 +417,190 @@ impl EthClient {
...
+    }
+
+    //  Verify POSA of the binance chain header.
+    fn verify_header_bsc(&self, header: &BlockHeader, prev: &BlockHeader) -> bool {
+        // The genesis block is the always valid dead-end
+        if header.number == 0 {
+            return true;
+        }
+
+        // verift basic header properties.
```
mfornet, line 463:
```suggestion
        // verify basic header properties.
```

mfornet (commented):
Contract LGTM.

For the BSC code in the client I think we should use [rust feature flags](https://doc.rust-lang.org/cargo/reference/features.html) so only the intended binary is compiled and deployed.

mfornet (approved):
Merge conflicts needs to be solved before merging.

idirall22:
Hello @mfornet @abacabadabacaba 
The conflicts are solved the fork is ready to be merged.

mfornet:
@sept-en could you take a look at this PR before merging.

PR #605, opened 2021-07-12
wip: verify key from state root and transaction root

vince0656:


vince0656:
WIP

vince0656:
@mfornet

contracts/near/eth-prover/src/lib.rs
```diff
@@ -149,6 +149,84 @@ impl EthProver {
...
+        #[serializer(borsh)] proof: Vec<Vec<u8>>,
+        #[serializer(borsh)] skip_bridge_call: bool,
+    ) -> PromiseOrValue<bool> {
+        self.check_not_paused(PAUSE_VERIFY);
+
+        let header: BlockHeader = rlp::decode(header_data.as_slice()).unwrap();
+
+        // Verify receipt included into header
+        let data =
+            Self::verify_trie_proof(header.state_root, state_key, proof);
```
mfornet, line 167: To proof some storage data is present at a fixed point, you need to go trough two tries. I'll refer to the [yellow paper](https://ethereum.github.io/yellowpaper/paper.pdf).

In the block header there is the state trie root [Section 4.1 from the YP]. This trie contains a mapping between address to the state of each account. In particular one of the fields for each account is the `storage_root`.

The `storage_root` is the root of yet another trie (there is one per account in ethereum). And this trie has the internal state of the account. Quoting from the YP:

```
storageRoot: A 256-bit hash of the root node of a Merkle Patricia tree 
that encodes the storage contents of the account 
(a mapping between 256-bit integer values), 
encoded into the trie as a mapping from the Keccak 256-bit hash of the 
256-bit integer keys to the RLP-encoded 256-bit integer values.
```

With that said, it means overall proving some data is in the state should look like this (there are some notes about the following code following it):

```python
def verify_state_entry(header, address, state_proof, storage_key, storage_proof):
    """
    header: Is the header where the proof is being executed. I.e the state must be present at this header. Usually it is passed encoded as RLP, so it must be decoded before start using it.
    address: The contract address where the state that is being checked lives.
    state_proof: Proof that allows building trustlessly the account state of previous address.
    storage_key: The key in the storage_trie of the data that is being checked. 
    storage_proof: Proof that allows building trustlessly the data that is being checked.
    """
    account_state_rlp = verify_trie_proof(header.state_root, address, state_proof)
    expected_storage_root = rlp.decode(account_state_rlp).storage_root
    _storage_data = verify_trie_proof(expected_storage_root, storage_key, storage_proof)
   
    # Check this is valid header by querying the Ethereum Light Client
    block_hash_safe(header.number).then(on_block_hash(header.hash))
```

1. Notice that we are actually not passing the `storage_data`. This is because the state data can be derived from the `storage_proof` easily. It is the data in the last step of the proof. This means, that when a contract builds the proof and call the prover to verify it, then if the call succeeds, the proof is valid, and so the data that was used to build this proof. This is the same reason why `_storage_data` is actually not used in the above code.

2. `verify_trie_proof` is a function that receives the expected trie root, the key, and the proof for this key. And it returns the data stored in this key under the trie. It will fail if the proof is invalid.

contracts/near/eth-prover/src/lib.rs
```diff
@@ -149,6 +149,84 @@ impl EthProver {
...
+            BLOCK_HASH_SAFE_GAS,
+        )
+            .then(remote_self::on_block_hash(
+                header.hash.unwrap(),
+                &env::current_account_id(),
+                0,
+                ON_BLOCK_HASH_GAS,
+            ))
+            .into()
+    }
```
mfornet, line 228: This function is looking good. However notice that we don't need the `tx_data` argument or the check `let verification_result = tx_data == data;`. This was explained in the previous comment. This is not how it is done for receipts, but that can be updated as well.

mfornet (commented):
It would be great if together with this PR we include the code to generate the proof both for tx and state (ideally in js/ts but could be in whatever other language).

This will be good to make tests on live data and to use it in our tooling later.

vince0656:
@mfornet I have added a JS script for getting a storage proof from a JSON RPC endpoint. One thing I am unsure of is how to generate the rlp for the header data? there is `eth_getBlockByNumber` that gets the block information easily but then is it just a case of rlp encoding that response?

I have tried to hard code the state root and supply the account proof from my script in a rust test but I am still getting assertion issues where it is not able to correctly verify the state root. Are you able to see if my new `lib.rs` code has any issues? I think it should work but it is not. I need to remove the bit about storage data but it's failing right at the first part of verifying the `state_root` trie i.e. here:

```
assert_eq!(near_keccak256(node), expected_root.as_slice());
```

contracts/near/eth-prover/src/lib.rs
vince0656: ok I believe I have addressed all points

contracts/near/eth-prover/src/lib.rs
vince0656: ok I believe I have addressed all points

PR #622, opened 2021-08-27
Migrate CI pipeline to GitHub actions

strokovok:
Right now all tests fail with same errors as on buildkite,
but after current errors fix another problems will probably appear.


sept-en:
@strokovok Some tests (myerc20) are failing in Github CI but are successful in Buildkite. It seems that maybe the local near node is being set up incorrectly. Could you please check?

PR #628, opened 2021-09-06
Create CODEOWNERS

sept-en:
Add @sept-en to CODEOWNERS.


mfornet (changes requested):
I think it is acceptable at this point to remove @abacabadabacaba from CODEOWNERS.

mfornet (approved):

Issue #616, opened 2021-08-10
initialize the near client "RlpIsTooShort"


idirall22:
Hello
### Error
I'm trying to run the bridge locally using the main branch, but I have an error related to block header RLP:

![gggg](https://user-images.githubusercontent.com/26064387/128828907-5409e69f-6d14-4244-a2d7-520f22aca796.PNG)

### Actions
- yarn
- yarn install
- cd contracts/eth/nearbridge/ && yarn && yarn build
- cd contracts/eth/nearprover/ && yarn && yarn build
- cli/index.js clean
- cli/index.js prepare
- cli/index.js start near-node
- cli/index.js start ganache
- cli/index.js init-near-contracts
- cli/index.js init-near-contracts  <---- ERROR

tests fail
![Capture](https://user-images.githubusercontent.com/26064387/130420415-6b35e02d-2792-4618-bd1b-3f0c827bfc38.PNG)

sept-en:
Fixed in #636

PR #636, opened 2021-10-14
Fix ci

karim-en:


karim-en:
Releated #616

cli/init/eth-contracts.js
```diff
@@ -1,6 +1,8 @@
 const BN = require('bn.js')
 const fs = require('fs')
-const { Web3, normalizeEthKey } = require('rainbow-bridge-utils')
+const { Web3, normalizeEthKey, sleep } = require('rainbow-bridge-utils')
+
+const RETRY_SEND_TX = 15
```
sept-en, line 5: Where does this constant come from?

testing/transfer-eth-erc20/from-near.js
```diff
@@ -299,19 +302,31 @@ class TransferEthERC20FromNear {
       const borshProofRes = borshifyOutcomeProof(proofRes)
       clientBlockHeight = new BN(clientBlockHeight)
       // Debugging output, uncomment for debugging.
-      // console.log(`proof: ${JSON.stringify(proofRes)}`);
-      // console.log(`client height: ${clientBlockHeight.toString()}`);
-      // console.log(`root: ${clientBlockMerkleRoot}`);
+      // console.log(`proof: ${JSON.stringify(proofRes)}`)
+      // console.log(`client height: ${clientBlockHeight}`)
```
sept-en, line 306: Remove commented-out code

testing/transfer-eth-erc20/from-near.js
```diff
@@ -299,19 +302,31 @@ class TransferEthERC20FromNear {
...
+      // TODO investigate and fix this web3 call
+      // const ethERC20Contract = new web3.eth.Contract(
+      //   JSON.parse(fs.readFileSync(ethErc20AbiPath)),
+      //   ethErc20Address,
+      //   {
+      //     from: ethMasterAccount,
+      //     handleRevert: true
+      //   }
+      // )
+      // const oldBalance = await ethERC20Contract.methods.balanceOf(ethReceiverAddress).call()
```
sept-en, line 321: Remove commented-out code

testing/transfer-eth-erc20/from-near.js
```diff
@@ -323,9 +338,9 @@ class TransferEthERC20FromNear {
           gasPrice: new BN(await robustWeb3.web3.eth.getGasPrice()).mul(new BN(ethGasMultiplier))
         }
       )
-      const newBalance = await ethERC20Contract.methods
-        .balanceOf(ethReceiverAddress)
-        .call()
+      // TODO investigate and fix this web3 call
+      // const newBalance = await ethERC20Contract.methods.balanceOf(ethReceiverAddress).call()
```
sept-en, line 342: What's about this TODO?

sept-en (approved):
Added some minor comments/questions, but others than that - LGTM! Thanks for fixing this!

cli/init/eth-contracts.js
karim-en: It is just a magic number like [this](https://github.com/aurora-is-near/rainbow-bridge/blob/master/utils/robust.js#L8) and [this](https://github.com/aurora-is-near/rainbow-bridge/blob/fd7019ff001968105e8bb22627f5a5097ba345a8/utils/helpers.js#L5) to get around of nonce error.

testing/transfer-eth-erc20/from-near.js
karim-en: There is an unknown reason why this call fail when loading the contract from the abi, I think this research is not urgent and could be fixed in a follow-up issue. So for now at least we should have a passed ci to detect buggy changes.

sept-en: Anyway, I think the `ethers` interface that you've used is better than loading the whole abi. Partial ABI is better, so please just remove this TODO then

PR #639, opened 2021-10-25
Add support for erc721 

karim-en:


mfornet:
We currently have too many CLI which we are maintaining for different use cases. In this particular case, I'd avoid having ERC20 + ERC721 connector logic outside of this repository. 

Let's discuss what will be the strategy ahead for maintaining a single CLI instead of keep all this info spread in several repositories @sept-en @karim-en

PR #646, opened 2021-11-29
Upgrade to nearcore 1.23.0

### LINKED ISSUE #644: Update Near Light Client to support new ValidatorStake
[ProtocolFeatures::BlockHeaderV3](https://github.com/near/nearcore/blob/master/core/primitives/src/version.rs#L174) on NEAR will change the way [next block producers are serialised](https://github.com/near/nearcore/blob/351519344cbe205a255dd6c1c2f0ae4a3f7c84d7/chain/chain/src/chain.rs#L506-L520). Notice the difference between [ValidatorStakeV1](https://github.com/near/nearcore/blob/52e822d08c514b1f846037e5e40032f2f2bbaee7/core/primitives/src/types.rs#L683-L690) and [ValidatorStakeV2](https://github.com/near/nearcore/blob/52e822d08c514b1f846037e5e40032f2f2bbaee7/core/primitives/src/types.rs#L483-L493). (One byte at the end indicating if it is a Chunk Producer or not).

Because of this the NearLightClient implementation on Ethereum should be updated.
- The [hash computation](https://github.com/aurora-is-near/rainbow-bridge/blob/63fb73ba3c3ced07b57c3040a5c6cdaff5d0f255/contracts/eth/nearbridge/contracts/NearDecoder.sol#L64) is correct and requires no changes, since it takes into account all the raw data.
- The way [block producer are deserialised](https://github.com/aurora-is-near/rainbow-bridge/blob/63fb73ba3c3ced07b57c3040a5c6cdaff5d0f255/contracts/eth/nearbridge/contracts/NearDecoder.sol#L35-L39) requires to be updated, taking into account the new field (one byte). The content of this byte can be ignored in the same way the account id is ignored (only used to compute the hash).

If the bridge is updated before this feature becomes live, we must make sure the bridge light client is aware about the protocol version, and deserialise each value according to the protocol version. Right now I'm investigating if this is possible.

---

Since only the Light Client needs to be upgraded, this can be done safely, and after the deployment is done properly, then the Prover can point to the new Light Client.

sept-en:
## Motivation
Close issue #644 (check issue for details)

## Changes

 `NearDecoder`: upgrade to `nearcore` 1.23.0:
* Handle enum byte for `ValidatorStake` version.
* `BlockProducer`: add `isChunkOnly` field.

Near2Eth-relay:
* utils/borsh.js: add support for optional `validator_stake_struct_version` field
  when borshifying light client block.

## Additional resources

You can check the [SSCCE](https://github.com/sept-en/borsh-validators-example/blob/master/src/lib.rs) verifying the change of the borsh-serialized result after wrapping the structure into an enum. 

## Tests

There are some tests that are failing due to this upgrade. Those tests rely on the old blocks which don't contain the enum for `ValidatorStake` version thus their hash will differ from the hash that will be computed having new rules with versioning (nearcore v1.23.0 upgrade adds an extra byte as a prefix to block producer which changes the hash). We can't change only `next_bp_hash` in files containing the blocks for tests as in this case signatures of validators will mismatch. The possible solutions to resolve tests:

1) Add another contract `NearBridgeV1` that will contain the old schema for decoding blocks prior to nearcore v1.23.0.
2) Add the possibility to compute the signature for older blocks correctly in `NearDecoder`.
2) Wait until the nearcore v1.23.0 is live on Testnet and after that use real Testnet blocks to update all failing tests.

**Option 1:** IMO is not really useful as we will test against old code-base that will not be supported anymore. Still, this will give us some insights if some tool is broken (e.g. near2eth-relay). But nevertheless, new tests for after-upgrade blocks should be added. So this looks like a redundant step.
**Option 2:** this will add another branch, increase the complexity of the code and this means that we are supporting an obsolete code. Also, this will include additional gas costs because the signature won't be computed from the single blob anymore.

So having all these things in mind, I'm for the **3rd** option but would like to hear any suggestions/ideas.


birchmd (approved):
LGTM

I also like option 3 for fixing the tests.

contracts/eth/nearbridge/contracts/NearDecoder.sol
```diff
@@ -7,6 +7,9 @@ library NearDecoder {
     using Borsh for Borsh.Data;
     using NearDecoder for Borsh.Data;
 
+    uint8 constant VALIDATOR_V1 = 1;
```
karim-en, line 10: As I see this variable is not used, could we avoid that?

karim-en (commented):
Looks good for me, I've just one minor comment.
BTW, are you sure that we can merge it with master? maybe we should create a separate branch for that and merge it with master when the update will be available on the mainnet

contracts/eth/nearbridge/contracts/NearDecoder.sol
sept-en: I'd prefer to keep the constant for the visibility so it will be easier to remember later how these constants for versions are produced and also that `VALIDATOR_V2` value is not just a random number.

sept-en:
@karim-en IMO it's okay to merge in `master`. The version that will be deployed later will have its own separate release.

karim-en (approved):

sept-en:
Note: it seems that it's possible to have up to 200 validators with this update. So we need to raise the limit to 100 too. Though this will increase gas fees for sure.

birchmd (approved):

contracts/eth/nearbridge/test/NearBridge3.js
```diff
@@ -0,0 +1,27 @@
...
+        ethers.BigNumber.from("20000000000"), // replace duration
+        await (await ethers.getSigners())[0].getAddress(),
+        0
+    );
+    await NearBridge.deposit({ value: ethers.utils.parseEther('1') });
+});
+
+it('Verify init with validators V1', async function () {
+    const initialValidators = borshifyInitialValidators(require('./block_validators_v1_testnet.json').next_bps);
+    console.log(`Initial validators: ${initialValidators.toString('hex')}`);
```
0x3bfc, line 23: Do we need this `console.log` ?

contracts/eth/nearbridge/contracts/NearDecoder.sol
```diff
@@ -30,12 +33,20 @@ library NearDecoder {
...
+        bool isChunkOnly;
     }
 
     function decodeBlockProducer(Borsh.Data memory data) internal pure returns (BlockProducer memory res) {
+        uint8 validator_version = data.decodeU8();
         data.skipBytes();
         res.publicKey = data.decodePublicKey();
         res.stake = data.decodeU128();
+        if (validator_version == VALIDATOR_V2) {
+            res.isChunkOnly = data.decodeU8() != 0;
```
0x3bfc, line 46: `res.isChunkOnly = data.decodeU8() != 0;` This line will always produce `true`.

0x3bfc (commented):
LGTM, I've left minor comments.

contracts/eth/nearbridge/contracts/NearDecoder.sol
sept-en: why?

contracts/eth/nearbridge/test/NearBridge3.js
sept-en: This was useful for the time of writing but I think could be removed, agree.

contracts/eth/nearbridge/contracts/NearDecoder.sol
```diff
@@ -7,6 +7,9 @@ library NearDecoder {
     using Borsh for Borsh.Data;
     using NearDecoder for Borsh.Data;
 
+    uint8 constant VALIDATOR_V1 = 0;
+    uint8 constant VALIDATOR_V2 = 1;
```
mfornet, line 11: FYI: When this issue is solved (https://github.com/near/nearcore/issues/5471) we should be using the PROTOCOL_VERSION (from the block) rather than this hardcoded version. Not action item here, just a comment.

utils/borsh.js
```diff
@@ -404,6 +407,9 @@ function borshify (block) {
     Buffer.concat(
       block.next_bps.map((nextBp) =>
         Buffer.concat([
+          nextBp.hasOwnProperty("validator_stake_struct_version") && nextBp.validator_stake_struct_version === 'V2'
```
mfornet, line 410: Is the field `validator_stake_struct_version` introduced in the new version?

mfornet (approved):

contracts/eth/nearbridge/test/block_validators_v1_testnet.json
```diff
@@ -0,0 +1 @@
...8RrrHTm","stake":"546282177598469747960331624999"},{"validator_stake_struct_version":"V1","account_id":"fortress.pool.f863973.m0","public_key":"ed25519:veXPT5KHyEgwmDqV41iFX9oZ3rH1ELUwqWtocAA7WC8","stake":"440120917627900576755211817069"},{"validator_stake_struct_version":"V1","account_id":"blockngine.pool.f863973.m0","public_key":"ed25519:CZrTtCP6XkkxWtr3ATnXE8FL6bcG5cHcxfmdRgN7Lm7m","stake":"485047708849773266720539666408"},{"validator_stake_struct_version":"V1","account_id":"shardlabs.pool.f863973.m0","public_key":"ed25519:DxmhGQZ6oqdxw7qGBvzLuBzE6XQjEh67hk5tt66vhLqL","stake":"480621524322218101752022303236"},{"validator_stake_struct_version":"V1","account_id":"stakingpower.pool.f863973.m0","public_key":"ed25519:HsXvTEyjhmitcdK1Uk6do6rzRqC7U3vtDk5CGCcHUrEn","stake":"471871567154659041926854422200"},{"validator_stake_struct_version":"V1","account_id":"lavenderfive.pool.f863973.m0","public_key":"ed25519:AzwAiLDqprZKpDjhsH7dfyvFdfSasmPTjuJUAHfX1Pg4","stake":"445374847495594702756596983569"},{"validator_stake_struct_version":"V1","account_id":"infiniteloop.pool.f863973.m0","public_key":"ed25519:2fbiLqksH5viWXYoteyfKP9qQawkRKw4YogRFcvG3Z7f","stake":"488482402962925655210618680533"}]}
```
joshuajbouw, line 1: Does this need to be all one line? Makes it a pain to read.

contracts/eth/nearbridge/contracts/NearDecoder.sol
joshuajbouw, line 46: Can't you use `validator_version` here instead?

joshuajbouw: If the data is not 0, return true. If it is 0, return false. How can that always produce true?

joshuajbouw (changes requested):

contracts/eth/nearbridge/contracts/NearDecoder.sol
sept-en: Yes, I saw that. But I don't expect this will happen soon.

utils/borsh.js
sept-en: Yes

contracts/eth/nearbridge/test/block_validators_v1_testnet.json
sept-en: This is test data. I don't think it's intended to read it :) I don't think you can verify anything just by reading it.

contracts/eth/nearbridge/contracts/NearDecoder.sol
sept-en: We need to extract it from borsh-data (changing the cursor to be precise). Otherwise, we will have some data left.

sept-en: @joshuajbouw yeah, I'm also confused how this could be always `true`. Maybe I'm missing something.
@0x3bfc please comment

0x3bfc: Sorry I was confused with the value of `data.decodeU8()`.

0x3bfc (approved):

PR #647, opened 2021-12-02
NearProver:add scripts&tests for bridge upgradability

sept-en:


contracts/eth/nearprover/utils/upgrade_bridge_address.js
```diff
@@ -0,0 +1,53 @@
...
+    const nearProverFactory = await ethers.getContractFactory('NearProver');
+    const nearProver = nearProverFactory.attach(proverAddress);
+
+    console.log(`Got prover at address: ${proverAddress}`);
+
+    const initialBridgeAddress = await nearProver.bridge();
+    console.log(`Initial bridge address: ${initialBridgeAddress}`);
+    console.log(`Trying to upgrade bridge address to: ${newBridgeAddress}`);
+
+    const adminWallet = new ethers.Wallet(process.env.ETH_PRIVATE_KEY, provider);
```
mfornet, line 18: We should consider allowing hardware wallet, or offline signing in this step, to avoid compromising the key during the upgrade process.

contracts/eth/nearprover/utils/upgrade_bridge_address.js
```diff
@@ -0,0 +1,53 @@
...
+        await nearProver.admin(),
+        "The used account is not an admin of NearProver"
+    );
+
+    // Mask matches only on the latest 20 bytes (to store the address)
+    const mask = ethers.BigNumber.from("0x000000000000000000000000ffffffffffffffffffffffffffffffffffffffff");
+    const response = await nearProver
+        .connect(adminWallet)
+        .adminSstoreWithMask(BRIDGE_ADDRESS_SLOT, newBridgeAddress, mask);
+    await response.wait().then(function(receipt) {
```
mfornet, line 31: Ethers by default only waits for 1 block, which is fine in most cases. In this sensitive case, let's use a bigger number (bigger than 5 at least).

```suggestion
    await response.wait(10).then(function(receipt) {
```

contracts/eth/nearprover/utils/upgrade_bridge_address.js
sept-en: Sure, this is already WIP and will be added really soon.

contracts/eth/nearprover/UPGRADE.md
```diff
@@ -0,0 +1,35 @@
...
+
+1. Create `.env` file inside `eth-custodian` directory: `$ touch .env`.
+
+2. Add to the file your Infura API key:
+`$ echo "INFURA_API_KEY=YOUR_INFURA_API_KEY_HERE" >> .env` <br/>
+
+3. Add to the file Ethereum Private key:
+`$ echo "ETH_PRIVATE_KEY=YOUR_ETH_PRIVATE_KEY_HERE" >> .env`
+
+## Mock deployment
```
0x3bfc, line 15: I would like to introduce some extra steps here as follow: 
- Integrating [slither](https://github.com/crytic/slither) as a smart contract static analysis tool.
- Apply the [upgradeability-check](https://github.com/crytic/slither/wiki/Upgradeability-Checks) detector.
- Finally logging the analysis results in a report attached to that upgrade.

This is just preventive mechanism against future upgradeability failure.

sept-en: @0x3bfc we are not talking about the upgradability as a proxy here. Here we are talking about changing the reference to the light client inside the existing smart contract using `AdminControlled` interface. So I agree we need `slither` but `upgradability-check` detector is not relevant for this specific PR.

0x3bfc: It make sense .... Thanks @sept-en 👍

contracts/eth/nearprover/utils/upgrade_bridge_address.js
```diff
@@ -0,0 +1,53 @@
...
+
+    // Mask matches only on the latest 20 bytes (to store the address)
+    const mask = ethers.BigNumber.from("0x000000000000000000000000ffffffffffffffffffffffffffffffffffffffff");
+    const response = await nearProver
+        .connect(adminWallet)
+        .adminSstoreWithMask(BRIDGE_ADDRESS_SLOT, newBridgeAddress, mask);
+    await response.wait(10).then(function(receipt) {
+        console.log(`Transaction mined: `);
+        console.log(receipt);
+    });
```
mfornet, line 34:
```suggestion
    let receipt = await response.wait(10);
    console.log('Transaction mined: ');
    console.log(receipt);
```

mfornet: Remember that 10 is the number of blocks so this is over 2 minutes (~140s)

contracts/eth/nearprover/hardhat.config.js
```diff
@@ -1,6 +1,28 @@
...
+const ETH_PRIVATE_KEY = process.env.ETH_PRIVATE_KEY;
+const INFURA_API_KEY = process.env.INFURA_API_KEY || '';
+
+task('upgrade-provers-bridge-address-to', 'Upgrades the provided prover to use the bridge at the provided address')
+    .addParam('prover', 'Prover address')
+    .addParam('newBridge', 'The address of the new bridge')
+    .setAction(async taskArgs => {
+        const { upgradeProversBridgeAddressTo } = require('./utils/upgrade_bridge_address.js');
+        await upgradeProversBridgeAddressTo(hre.ethers.provider, taskArgs.prover, taskArgs.newBridge);
+    });
```
mfornet, line 15: I'm used to use typescript for this, so the code looks a little bit weird. My editor is complaining that `task` and `hre` are not declared. In TS it is done the following way:

```suggestion
    .setAction(async (taskArgs, hre) => {
        const { upgradeProversBridgeAddressTo } = require('./utils/upgrade_bridge_address.js');
        await upgradeProversBridgeAddressTo(hre.ethers.provider, taskArgs.prover, taskArgs.newBridge);
    });
```

sept-en: @mfornet this is works because this script is being run with a hardhat command which provides all that needed env data. But yes, I agree it's better to have it suitable to run with the default commands too

contracts/eth/nearprover/utils/upgrade_bridge_address.js
sept-en: It is, though if the receipt is available earlier, it will return the receipt immediately.

mfornet: Interesting, I thought it was waiting for 10 blocks on top of the block where the receipt was included.

0x3bfc (approved):
LGTM

contracts/eth/nearprover/hardhat.config.js
```diff
@@ -13,5 +35,23 @@ module.exports = {
                 runs: 1000
             }
         }
+    },
+    networks: {
+        ropsten: {
+            url: `https://ropsten.infura.io/v3/${INFURA_API_KEY}`,
+            accounts: [`0x${ETH_PRIVATE_KEY}`],
```
karim-en, line 42: Please add an explicit note in the readme that the `ETH_PRIVATE_KEY` should be set without '0x'

contracts/eth/nearprover/UPGRADE.md
```diff
@@ -0,0 +1,35 @@
+# Upgradability
+
+This file describes how to upgrade the bridge address in NearProver contract.
+
+## Configuration
+
+1. Create `.env` file inside `eth-custodian` directory: `$ touch .env`.
```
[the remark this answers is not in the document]
sept-en: already fixed in #651

contracts/eth/nearprover/hardhat.config.js
sept-en: Fixed in f5183d1

mfornet (approved):

karim-en (approved):

PR #650, opened 2021-12-14
NearOnEthClinet: add message for BP num limit

sept-en:
* Add an explicit message if the number of block producers for the provided block is bigger than `MAX_BLOCK_PRODUCERS`

Without the message, it becomes really hard to debug if something goes wrong.

NB: as Near Testnet currently allows having >100 Validators, it makes sense to consider changing the number of `MAX_BLOCK_PRODUCERS` to something meaningful in the future (maybe 256?)


mfornet (approved):
We should consider in the future having some sort of simple structured errors. This will help for consistency, and for parsing errors manually through our tools.

For example in this case it could be something like: `ERR_MAX_BLOCK_PRODUCERS: It is not expected having that many block producers for the provided block`

karim-en (approved):

PR #662, opened 2021-12-27
fix a bug validate the epoch header using the previous validator set

idirall22:
Hello Guys!!
I fixed a 1 million bounty bug 😂😂
 
This is a fix for a bug that happens on the epoch headers.
before this fix, the epoch headers was validated with the new validator set but this makes the validation fails on some epochs (1750200, 11693400, 12819400).

you can see [here](https://github.com/binance-chain/bsc/blob/master/consensus/parlia/parlia.go#L553) that they use the previous snapshot to validate the new header.

### DIFF
--- a/contracts/near/eth-client/src/lib.rs
+++ b/contracts/near/eth-client/src/lib.rs
@@ -574,14 +574,10 @@ impl EthClient {
 
     // check if the author address is valid and is in the validator set.
     #[cfg(feature = "bsc")]
-    fn bsc_is_validator(&self, header: &BlockHeader) -> bool {
+    pub fn bsc_is_validator(&self, header: &BlockHeader) -> bool {
         let (extra_vanity, extra_seal, address_size) = (32, 65, 20);
 
-        let epoch_header = if EthClient::is_epoch(header.number) {
-            header.clone()
-        } else {
-            self.bsc_get_epoch_header()
-        };
+        let epoch_header = self.bsc_get_epoch_header();
 
         if !self.bsc_is_in_validator_set(&epoch_header, header.author) {
             return false;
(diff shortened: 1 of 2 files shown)


contracts/near/eth-client/src/lib.rs
```diff
@@ -574,14 +574,10 @@ impl EthClient {
 
     // check if the author address is valid and is in the validator set.
     #[cfg(feature = "bsc")]
-    fn bsc_is_validator(&self, header: &BlockHeader) -> bool {
+    pub fn bsc_is_validator(&self, header: &BlockHeader) -> bool {
```
karim-en, line 577: This is a bad practice to make the function pub just for the tests.
Please turn it back and find a workaround
Refs:
https://doc.rust-lang.org/book/ch11-03-test-organization.html#testing-private-functions
https://stackoverflow.com/questions/27978377/how-do-i-test-private-methods-in-rust

mfornet (approved):
Good catch. I wonder if there is any good way to fuzz test this code and try to catch more errors like this. Non trivial of course, given that this scenarios happens with low probability.

contracts/near/eth-client/src/lib.rs
idirall22: Hey Karim!
I'll fix it thx :)

idirall22: Done

idirall22:
Hey Marcelo @mfornet 
The easiest way that I found is to implement a js script that validates the  epoch headers signer + difficulty (I ran the script between epoch 11000000 and 12000000 on mainnet)

karim-en (approved):

sept-en (approved):

PR #668, opened 2021-12-29
Upgrade admin: minor improvements

karim-en:
* Ignore case for admin address
* Log the mask and selected slot content
* Update README

### DIFF
--- a/contracts/eth/nearprover/utils/upgrade_admin.js
+++ b/contracts/eth/nearprover/utils/upgrade_admin.js
@@ -20,10 +20,11 @@ async function upgradeAdminAddressTo ({
     adminAddressSlot,
     ledgerKeyPath,
 }) {
-    const currentAdminAtSlot = ethers.BigNumber.from(await provider.getStorageAt(contractAddress, Number(adminAddressSlot))).toHexString();
+    const slotContent = await provider.getStorageAt(contractAddress, Number(adminAddressSlot));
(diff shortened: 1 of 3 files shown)


contracts/eth/nearprover/utils/upgrade_admin.js
```diff
@@ -20,10 +20,11 @@ async function upgradeAdminAddressTo ({
     adminAddressSlot,
     ledgerKeyPath,
 }) {
-    const currentAdminAtSlot = ethers.BigNumber.from(await provider.getStorageAt(contractAddress, Number(adminAddressSlot))).toHexString();
+    const slotContent = await provider.getStorageAt(contractAddress, Number(adminAddressSlot));
+    const currentAdminAtSlot = ethers.BigNumber.from(slotContent).toHexString();
```
sept-en, line 24: it's just slot content, we shouldn't expect by default that there will be an admin, in this case, this is exactly the purpose of all those checks. We rather should signalize that admin is not in the selected slot and the current slot content is the following (printing the content)

sept-en (approved):
Added minor comment, other than that LGTM!

contracts/eth/nearprover/utils/upgrade_admin.js
karim-en: fixed at 86fbac9f5b3286500a4206fd795e1cb695494740, please check

Issue #669, opened 2022-01-04
BSC: Improve test coverage


karim-en:
The negative tests are missing, we need more tests.

PR #670, opened 2022-01-04
BSC: add unit tests

karim-en:
Issue #669
Were added 3 tests:

1. Negative test  of adding  blocks [0,10] : success
2. Negative test with invalid block number: success
3. Negative test with invalid gas limit: success 
4. Positive test for gc action: success


contracts/near/eth-client/src/tests.rs
```diff
@@ -574,6 +574,89 @@ fn bsc_validate_headers() {
...
+        None,
+        chain_id,
+    );
+
+    for block in blocks.into_iter().skip(1) {
+        contract.add_block_header(block, vec![]);
+    }
+    assert!(hashes[0] == contract.epoch_header);
+    assert!(contract.headers.len() == 10);
+}
```
sept-en, line 602: This is not relevant anymore after #679, this should be either negative or removed

sept-en: Fixed in e6e88c

idirall22:
LGTM

Issue #672, opened 2022-01-05
BSC: any block with number=0 ignored from validation


karim-en:
It is possible to add any block to the client just by changing the block number to 0, see the [test](https://github.com/aurora-is-near/rainbow-bridge/blob/0b13aa89bcdfa1975e5e62f6e62883deffeff622/contracts/near/eth-client/src/tests.rs#L607)

Alleged bug at this [line](https://github.com/aurora-is-near/rainbow-bridge/blob/bsc/contracts/near/eth-client/src/lib.rs#L474)

karim-en:
@idirall22 FYI

idirall22:
No, s the block epoch (number % 200 == 0) contains the validator set, you have to submit the blocks in sequence one after one. So; the first block should always be an epoch block and the next one have to be block number +1

karim-en:
> No, s the block epoch (number % 200 == 0) contains the validator set, you have to submit the blocks in sequence one after one. So; the first block should always be an epoch block and the next one have to be block number +1
@idirall22 ok, but what about this line https://github.com/aurora-is-near/rainbow-bridge/blob/bsc/contracts/near/eth-client/src/lib.rs#L474 and this [test](https://github.com/aurora-is-near/rainbow-bridge/blob/0b13aa89bcdfa1975e5e62f6e62883deffeff622/contracts/near/eth-client/src/tests.rs#L607) 

Are you sure that this line is safe? if yes, then please explain why
```
        if header.number == 0 {
            return true;
        }
```

idirall22:
@karim-en it's related to the parlia (BSC) consensus implementation see [here](https://github.com/binance-chain/bsc/blob/master/consensus/parlia/parlia.go#L548)

karim-en:
@idirall22 in the link which you provided the function returns the error `errUnknownBlock` but our `bsc_verify_header` returns true, that is what is confusing me is the validation step.
But what if we turn `return true` to `return false` and just prevent adding blocks with `number = 0`, is there anything preventing us to do that?

idirall22:
@karim-en
I made a mistake when I copied the line from Github see [here](https://github.com/binance-chain/bsc/blob/master/consensus/parlia/parlia.go#L383)
in the Parlia consensus they are checking for the block number == 0 two times
1st time they return nil which mean there is no error (in our case true) the second check we don't need to do add it as the first one is enough.

PR #679, opened 2022-01-07
BSC client: remove zero block trust

sept-en:
This rule doesn't produce any useful impact as we don't suppose to run the light-client from the genesis block. Meantime, the rule is dangerous from the security point of view as it relies on further validation in other parts of the code.

Related #672.


karim-en (approved):

karim-en (approved):

idirall22 (approved):
Yes, it make sense
LGTM

mfornet (approved):

PR #683, opened 2022-01-08
EthOnNearClient: fix pre-London block deserialization

### LINKED ISSUE #682: Broken EthOnNear client tests after EIP-1559 update
Some tests for Eth on NEAR `eth-client` are broken after fe6ed9e commit. This is related to the EIP-1559 update and requires fixing it.

An example of failure:

```
thread 'tests::add_dags_merkle_roots' panicked at 'called `Result::unwrap()` on an `Err` value: RlpIsTooShort', eth-client/src/lib.rs:124:72                                                                                                
stack backtrace:                                                                                                                                                                                                                            
   0: rust_begin_unwind                                                                                                                                                                                                                     
             at /rustc/../library/std/src/panicking.rs:483                                                                                                                                            
   1: core::panicking::panic_fmt                                                                                                                                                                                                            
             at /rustc/../library/core/src/panicking.rs:85                                                                                                                                            
   2: core::option::expect_none_failed                                                                                                                                                                                                      
             at /rustc/../library/core/src/option.rs:1234                                                                                                                                             
   3: core::result::Result<T,E>::unwrap                                                                                                                                                                                                     
             at /rustc/../library/core/src/result.rs:973                                                                                                                                              
   4: eth_client::EthClient::init                                                                                                                                                                                                           
             at ./src/lib.rs:124                                                                                                                                                                                                            
   5: eth_client::tests::add_dags_merkle_roots                                                                                                                                                                                              
             at ./src/tests.rs:250                                                                                                                                                                                                          
   6: eth_client::tests::add_dags_merkle_roots::{{closure}}                                                                                                                                                                                 
             at ./src/tests.rs:245                                                                                                                                                                                                          
   7: core::ops::function::FnOnce::call_once                                                                                                                                                                                                
             at /rustc/../library/core/src/ops/function.rs:227                                                                                                                                        
   8: core::ops::function::FnOnce::call_once                                                                                                                                                                                                
             at /rustc/../library/core/src/ops/function.rs:227
```

sept-en:
Summary:
* EthOnNearClient: fix pre-London block deserialization 
* Don't run pre-London tests when the EIP-1559 feature is active.
* Run both test suites: with and without default features (with and
  without EIP-1559 support).

The issue happened as `BorshDeserialize::deserialize()` modifies the buffer so if the attempt to deserialize to London black fails, the buffer is already updated and it can not be deserialized as a pre-London block anymore. However, I don't really like that approach as it requires twice-deserialization for the London blocks. We can manually shift the buffer if the deserialization of the London block is successful too.

Fixes #682.

### DIFF
--- a/contracts/near/eth-types/src/lib.rs
+++ b/contracts/near/eth-types/src/lib.rs
@@ -299,8 +299,8 @@ impl From<BlockHeaderPreLondon> for BlockHeader {
 
 impl BorshDeserialize for BlockHeader {
     fn deserialize(buf: &mut &[u8]) -> std::io::Result<Self> {
-        if let Ok(header) = BlockHeaderLondon::deserialize(buf) {
-            Ok(header.into())
+        if let Ok(_header) = BlockHeaderLondon::try_from_slice(buf) {
+            BlockHeaderLondon::deserialize(buf).map(Into::into)
         } else {
             BlockHeaderPreLondon::deserialize(buf).map(Into::into)
         }
--- a/contracts/near/eth-client/src/tests.rs
+++ b/contracts/near/eth-client/src/tests.rs
@@ -242,6 +242,7 @@ fn assert_hashes_equal_to_contract_hashes(contract: &EthClient, heights: &[u64],
 }
 
 #[test]
+#[cfg_attr(feature = "eip1559", ignore)]
 fn add_dags_merkle_roots() {
     testing_env!(get_context(vec![], false));
     let (blocks, _) = get_blocks(&WEB3RS, 400_000, 400_001);
@@ -267,6 +268,7 @@ fn add_dags_merkle_roots() {
 }
 
 #[test]
+#[cfg_attr(feature = "eip1559", ignore)]
 fn add_blocks_2_and_3() {
     testing_env!(get_context(vec![], false));
 
@@ -305,6 +307,7 @@ fn add_blocks_2_and_3() {
 }
 
 #[test]
+#[cfg_attr(feature = "eip1559", ignore)]
 fn add_blocks_before_and_after_istanbul_fork() {
     testing_env!(get_context(vec![], false));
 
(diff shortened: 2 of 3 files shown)


contracts/near/eth-types/src/lib.rs
```diff
@@ -299,8 +299,8 @@ impl From<BlockHeaderPreLondon> for BlockHeader {
 
 impl BorshDeserialize for BlockHeader {
     fn deserialize(buf: &mut &[u8]) -> std::io::Result<Self> {
-        if let Ok(header) = BlockHeaderLondon::deserialize(buf) {
-            Ok(header.into())
+        if let Ok(_header) = BlockHeaderLondon::try_from_slice(buf) {
+            BlockHeaderLondon::deserialize(buf).map(Into::into)
```
mfornet, line 303: You can clone buf pointer, since the mutable part is the pointer itself but not the underlying data.
For the second branch of the if, you can use the cloned buffer, that way it is not required to deserialise twice.

mfornet (changes requested):

contracts/near/eth-types/src/lib.rs
mfornet:
```rs
impl BorshDeserialize for BlockHeader {
    fn deserialize(buf: &mut &[u8]) -> std::io::Result<Self> {
        let buf1 = &mut buf.clone();

        if let Ok(header) = BlockHeaderLondon::try_from_slice(buf) {
            Ok(header.into())
        } else {
            BlockHeaderPreLondon::deserialize(buf1).map(Into::into)
        }
    }
}
```

mfornet (changes requested):

karim-en (approved):

contracts/near/eth-types/src/lib.rs
sept-en: This specific piece of code isn't working. Moreover, for the first case (London block headers), we actually need to make sure that the pointer is modified as the header could be a part of some more complex structure. So if we don't modify the pointer, further borsh deserialization won't happen and it would be hard to debug that issue in the future.

mfornet: I figured out that there are collisions between both types of blocks! so maybe we should consider changing this data structure to an enum. The problem is that one can be a prefix of the other and use the wrong structure. Still need to investigate if this can be exploited in a significant way. 

Notice this is unrelated to changes made on the PR.

mfornet (approved):

karim-en (approved):

PR #687, opened 2022-01-11
CI: disable old tests prior to nearcore 1.23.0.

sept-en:
I'd want to disable those tests as they are no longer used and working after nearcore 1.23.0 updated. The motivation for not removing those tests completely: when the new tests (updated ones) are added, it will be easier to review and verify that those will be actually the same tests but with different input data.


karim-en (commented):
I agree that we should keep those tests for easier review in the feature, but I think it is better to skip those tests instead of commenting them out.

mfornet (approved):

karim-en (approved):

Issue #685, opened 2022-01-09
BSC: fix the validator set changing rules


sept-en:
Currently, the validator set is being changed in the BSConNear client at the beginning of every epoch. However, in BSC, the validator set changes after N/2 blocks (where N is the number of validators of the previous epoch). This was initially required for BSC light-client as a way to verify that the light-client is not going to the wrong chain if the validator misbehaves and sends the data off-chain (or some other way).

As we are maintaining the light-client too it's also mandatory for us and moreover, this is required because otherwise, we won't be able to correctly verify the blocks. I also believe this is the cause of #661 which was revealing itself on Testnet not that often only because validators on Testnet don't change that often as they could possibly do on Mainnet.

Issue #661, opened 2021-12-27
BSC Fail to validate the signer and difficult


idirall22:
Hello Guys!

We started the bsc bridge on the testnet, everything was good we validated more than 50k BSC blocks. But when we reached the epoch 15180200 the bsc consensus failed to validate the headers [here](https://github.com/aurora-is-near/rainbow-bridge/blob/bsc/contracts/near/eth-client/src/lib.rs#L609)

After debugging, we observed that only the first 4 blocks are not valid [ 15180200, 15180201, 15180202, 15180203 ], but the next epochs are all valid.

Also, we observed that this happens when the validator set increases:
- epoch 15180000  => 6 validators
- epoch 15180200  => 7 validators
There is no issue when the validator set decreases.
**The inturn is not equal to the block signer and the difficulty also is not valid**.

We believe that for testnet they are ignoring the signer and difficulty checks see [here](https://github.com/binance-chain/bsc/blob/master/consensus/parlia/parlia.go#L215)

bsc consensus [here](https://github.com/binance-chain/bsc/blob/master/consensus/parlia/parlia.go#L587).

sept-en:
@idirall22 we can close it because of #662, right?

idirall22:
@sept-en
No this is not related to epoch block #662, it's related to the block signer and difficulty, which is not yet fixed.

PR #688, opened 2022-01-11
fixed validator changing rules

idirall22:
I fixed the issue related to validator epoch changing rule
- Fixe for #661
- Fixe for #685


idirall22:
Hey @sept-en 
I opened a PR but I'll test the code on testnet when I get time this evening(I'm still on vacation :sweat_smile:  )
In the meantime it'll be good if any one can review it

idirall22:
@sept-en 
I tested the client on testnet and it works fine.

sept-en:
Hi @idirall22, thanks for your PR! Unfortunately, this will not work in a real situation when we might have several diverged chains. So I'm closing this in favor of #693 where all that stuff is handled already. I would appreciate if you can join and review it too :)

PR #702, opened 2022-01-18
BSC: fix error `Header is already known`

karim-en:
The relayer uses `pub fn last_block_number` to get the last submitted block, but this logic doesn't work for BSC client because it returns the last final block and this led to getting the error `Header is already known. Number` on submitting blocks.
In this pull request, we check if the block has already been submitted before submitting it.

### DIFF
--- a/eth2near/eth2near-block-relay/index.js
+++ b/eth2near/eth2near-block-relay/index.js
@@ -161,6 +161,18 @@ class Eth2NearRelay {
         }
       }
 
+      const isBlockAlreadySubmitted = async (chainBlock) => {
+        try {
+          const clientHashes = await this.ethClientContract.known_hashes(
+            chainBlock.number
+          )
+          return clientHashes.find((x) => x === chainBlock.hash)
+        } catch (e) {
+          console.error(e)
+          return await isBlockAlreadySubmitted(chainBlock)
+        }
+      }
+
       const estimatedValued = (previousBlockNumber === undefined) ? 0 : clientBlockNumber - (previousBlockNumber + this.totalSubmitBlock)
 
       /// In case there exist a fork, find how many steps should go backward (delta) to the first block
@@ -184,6 +196,12 @@ class Eth2NearRelay {
           for (let i = clientBlockNumber + 1; i <= endBlock; i++) {
             if (this.validateHeaderMode === 'bsc') {
               const block = await this.robustWeb3.getBlock(i)
+              if (await isBlockAlreadySubmitted(block)) {
+                if (endBlock <= chainBlockNumber) {
+                  endBlock++
+                  continue
+                }
+              }
               blockPromises.push({ header_rlp: block })
             } else {
               blockPromises.push(this.getParseBlock(i))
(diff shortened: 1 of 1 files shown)


eth2near/eth2near-block-relay/index.js
```diff
@@ -184,35 +196,42 @@ class Eth2NearRelay {
...
             } else {
               blockPromises.push(this.getParseBlock(i))
             }
           }
 
           const blocks = await Promise.all(blockPromises)
+          const firstBlock = blocks[0].header_rlp
           console.log(
-            `Got and parsed block ${clientBlockNumber + 1} to block ${endBlock}`
+            `Got and parsed block ${firstBlock.number} to block ${endBlock}`
```
sept-en, line 214: As you are doing some little refactoring here already, let's eliminate a possible confusion so `endBlock` should be `endBlockNumber`

eth2near/eth2near-block-relay/index.js
```diff
@@ -184,35 +196,42 @@ class Eth2NearRelay {
...
+            const action = this.submitBlock(block)
             actions.push(action)
           }
 
           const task = this.ethClientContract.account.signAndSendTransaction(this.ethClientContract.contractId, actions)
 
           console.log(
-            `Submit txn to add block ${clientBlockNumber + 1
+            `Submit txn to add block ${firstBlock.number
             } to block ${endBlock}`
```
sept-en, line 228: This should be a single line for sure

eth2near/eth2near-block-relay/index.js
```diff
@@ -184,35 +196,42 @@ class Eth2NearRelay {
           for (let i = clientBlockNumber + 1; i <= endBlock; i++) {
             if (this.validateHeaderMode === 'bsc') {
               const block = await this.robustWeb3.getBlock(i)
+              if (await isBlockAlreadySubmitted(block)) {
+                if (endBlock <= chainBlockNumber) {
+                  endBlock++
+                  continue
+                }
+              }
```
sept-en, line 204: Why do we need such error-prone (IMO) construction? I believe we can just have it like this:

```suggestion
              if (await isBlockAlreadySubmitted(block)) {
                continue
              }
```

sept-en (changes requested):

eth2near/eth2near-block-relay/index.js
```diff
@@ -161,6 +161,18 @@ class Eth2NearRelay {
...
 
+      const isBlockAlreadySubmitted = async (chainBlock) => {
+        try {
+          const clientHashes = await this.ethClientContract.known_hashes(
+            chainBlock.number
+          )
+          return clientHashes.find((x) => x === chainBlock.hash)
+        } catch (e) {
+          console.error(e)
+          return await isBlockAlreadySubmitted(chainBlock)
```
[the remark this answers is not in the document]
karim-en: This line has been added just to unify the error handling in this file. So from my point of view if this logic was reviewed and tested then we should reuse it. But if from your point of view this concept is not correct, then we should refactor all code that uses this logic.
PS: this logic was introduced here https://github.com/aurora-is-near/rainbow-bridge/commit/9bb698508ddc96c5cd25017596d3372b4690d3c3#diff-d27d0bd772a13324192947711180ae6f0b2d54e6008fb64d74f801a322baf629R158

eth2near/eth2near-block-relay/index.js
karim-en: Fixed 147a517f09f3adeada85f845260a40d60b2ba346

eth2near/eth2near-block-relay/index.js
karim-en: Fixed f3c20251c33750280ab677256c6c7f6da1f438fc

eth2near/eth2near-block-relay/index.js
karim-en: Changed f3c20251c33750280ab677256c6c7f6da1f438fc

eth2near/eth2near-block-relay/index.js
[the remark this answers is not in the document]
sept-en: Ok, agreed. Please create an appropriate issue for that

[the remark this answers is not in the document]
sept-en: Relevant issue: #703

sept-en (approved):

PR #707, opened 2022-01-31
Polygon Bridge, working in progress...

idirall22:
Polygon Bridge, working in progress...


sept-en (commented):
Closed PR as stalled. All future work, if relevant, should be continued in a separate PR.

PR #727, opened 2022-03-17
NearBridge: optimize Solidity slots packing

sept-en:


mfornet (approved):

karim-en (approved):
LGTM.
But maybe it would be good to use some solidity analysis tools to cover this and other optimizations and vulnerabilities

contracts/eth/nearbridge/contracts/NearBridge.sol
```diff
@@ -21,33 +21,35 @@ contract NearBridge is INearBridge, AdminControlled {
...
     }
 
-    // Whether the contract was initialized.
-    bool public initialized;
     uint256 public lockEthAmount;
     // lockDuration and replaceDuration shouldn't be extremely big, so adding them to an uint64 timestamp should not overflow uint256.
     uint256 public lockDuration;
     // replaceDuration is in nanoseconds, because it is a difference between NEAR timestamps.
     uint256 public replaceDuration;
     Ed25519 immutable edwards;
```
0x3bfc, line 29: Not sure if this the correct slot for `edwards`.

contracts/eth/nearbridge/contracts/NearBridge.sol
```diff
@@ -21,33 +21,35 @@ contract NearBridge is INearBridge, AdminControlled {
...
+    // Address of the account which submitted the last block.
+    address lastSubmitter;
+
+    // Whether the contract was initialized.
+    bool public initialized;
     bool untrustedNextEpoch;
     bytes32 untrustedHash;
     bytes32 untrustedMerkleRoot;
     bytes32 untrustedNextHash;
+    uint256 untrustedTimestamp;
```
0x3bfc, line 47: It is better to group all the variables with the same slot size (256) together. I know it is also confusing the state variables readability ... but in general it is ok

0x3bfc (approved):
Minor comments.

contracts/eth/nearbridge/contracts/NearBridge.sol
sept-en: yeah, I agree on that in general. In this specific case, as the contract is quite complex I strive more for having readability too and at least somehow group public/private fields. I know that this `initialized` is out of the rule here too, as an exception. As grouping uint256 makes no sense for the performance optimization, I'd leave it as-is for readability purposes.

contracts/eth/nearbridge/contracts/NearBridge.sol
sept-en: As it neither affects the performance nor the readability, I'd rather leave it as is.

PR #729, opened 2022-03-18
NearBridge: `challenge` method improvements

sept-en:
* `NearBridge`: `challenge()` - use `call()` instead of `transfer()`.
* `NearBridge`: `challenge()` method won't be reentrant because of `lastValidAt = 0` will happen before `call()`
* `NearProver`: fix dependency order in `package.json`.

### DIFF
--- a/contracts/eth/nearbridge/contracts/NearBridge.sol
+++ b/contracts/eth/nearbridge/contracts/NearBridge.sol
@@ -1,14 +1,12 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 pragma solidity ^0.8;
 
-import "@openzeppelin/contracts/security/ReentrancyGuard.sol";
-
 import "./AdminControlled.sol";
 import "./INearBridge.sol";
 import "./NearDecoder.sol";
 import "./Ed25519.sol";
 
-contract NearBridge is INearBridge, AdminControlled, ReentrancyGuard {
+contract NearBridge is INearBridge, AdminControlled {
     using Borsh for Borsh.Data;
     using NearDecoder for Borsh.Data;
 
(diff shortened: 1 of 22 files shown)


contracts/eth/nearbridge/contracts/NearBridge.sol
```diff
@@ -88,13 +92,18 @@ contract NearBridge is INearBridge, AdminControlled {
         payable(msg.sender).transfer(amount);
     }
 
-    function challenge(address payable receiver, uint signatureIndex) public override pausable(PAUSED_CHALLENGE) {
+    function challenge(address payable receiver, uint signatureIndex)
+        external
+        override
+        nonReentrant
```
mfornet, line 98: No need `nonReentrant`. It is acceptable at this point, since this function is not called often.

mfornet (approved):

contracts/eth/nearbridge/contracts/NearBridge.sol
```diff
@@ -1,12 +1,16 @@
...
 
+import "@openzeppelin/contracts/security/ReentrancyGuard.sol";
+
 import "./AdminControlled.sol";
 import "./INearBridge.sol";
 import "./NearDecoder.sol";
 import "./Ed25519.sol";
 
-contract NearBridge is INearBridge, AdminControlled {
+import "@openzeppelin/contracts/security/ReentrancyGuard.sol";
```
0x3bfc, line 11: imported twice!

0x3bfc (approved):
LGTM with a minor change.

karim-en (approved):

contracts/eth/nearbridge/contracts/NearBridge.sol
0x3bfc: Exactly because this state variable will block reentrancy by default: 

`lastValidAt = 0;`

karim-en (approved):

0x3bfc (approved):

PR #730, opened 2022-03-18
CI: Test downgrading nearup

strokovok:


sept-en:
NB: closed as `nearup` upstream was updated.

PR #735, opened 2022-04-14
NEAR2Eth-Relay: Optimized Block Selection

vldmkr:
### Summary:
- Added documentation for newcomers to get started developing NEAR2Eth-Relay.
- Added CLI argument `near2eth-relay-select-duration` that sets the number of seconds to select the optimal block to submit.
- Added CLI argument `eth-use-eip-1559` which allows submitting transactions using the EIP-1559 way.
- Added a state inside the main loop to hold the block during the optimal block selection algorithm.
- Implemented delay calculation for optimal block selection.
- RobustWeb3 is extended with `maxPriorityFeePerGas` and `getFeeData` methods.

Debugging was done locally on the Hardhat Network and remotely on the Rinkeby testnet. The results can be seen on Etherscan: [0xbabe69df0d9ca729a546fc2f8823660d3d641855](https://rinkeby.etherscan.io/address/0xbabe69df0d9ca729a546fc2f8823660d3d641855). The last few transactions were alternately formed with and without EIP-1559 for clarity.

### DIFF
--- a/near2eth/near2eth-block-relay/index.js
+++ b/near2eth/near2eth-block-relay/index.js
@@ -197,16 +197,20 @@ class Near2EthRelay {
     const clientHeightGauge = httpPrometheus.gauge('client_height', 'amount of block client processed')
     const chainHeightGauge = httpPrometheus.gauge('chain_height', 'current chain height')
 
-    let firstSeenBlockTimestamp = null
     const nextBlockSelection = {
+      startedAt: 0,
       borshBlock: null,
       height: 0,
       set: function ({ borshBlock, lightClientBlock }) {
+        if (this.isEmpty()) {
+          this.startedAt = lightClientBlock.inner_lite.timestamp
+        }
         this.borshBlock = borshBlock
         this.height = lightClientBlock.inner_lite.height
         console.log(`The new optimal block is found. Height: ${this.height}. Size: ${this.borshBlock.length} bytes`)
       },
       clean: function () {
+        this.startedAt = 0
         this.borshBlock = null
         this.height = 0
       },
@@ -255,7 +259,6 @@ class Near2EthRelay {
             .add(replaceDuration)
             .sub(web3.utils.toBN(lastBlock.inner_lite.timestamp))
         }
-        firstSeenBlockTimestamp = firstSeenBlockTimestamp || lastBlock.inner_lite.timestamp
         // console.log({bridgeState, currentBlockHash, lastBlock, replaceDuration}) // DEBUG
         if (bridgeState.currentHeight < lastBlock.inner_lite.height) {
           if (nextValidAt.isZero() || replaceDelay.cmpn(0) <= 0) {
@@ -273,17 +276,9 @@ class Near2EthRelay {
               nextBlockSelection.set(blockCouple)
             }
 
-            // Calculation of selection delay starting from the first block seen after service restart
-            let selectDelay = selectDuration
-              .add(web3.utils.toBN(firstSeenBlockTimestamp))
+            const selectDelay = selectDuration
+              .add(web3.utils.toBN(nextBlockSelection.startedAt))
               .sub(web3.utils.toBN(lastBlock.inner_lite.timestamp))
-            if (!nextValidAt.isZero()) {
-              // Make sure that every time after restarting the service we have a delay for block selection
-              selectDelay = web3.utils.BN.max(
-                selectDelay,
-                selectDuration.add(replaceDelay)
-              )
-            }
             if (selectDelay.cmpn(0) > 0) {
               const selectDelaySeconds = selectDelay.div(web3.utils.toBN(1000_000_000))
               console.log(`Last light client block: ${lastBlock.inner_lite.height}`)
--- a/near2eth/near2eth-block-relay/index.js
+++ b/near2eth/near2eth-block-relay/index.js
@@ -174,6 +174,7 @@ class Near2EthRelay {
     near2ethRelayMinDelay,
     near2ethRelayMaxDelay,
     near2ethRelayErrorDelay,
+    near2ethRelaySelectDuration,
     ethGasMultiplier
   }) {
     const clientContract = this.clientContract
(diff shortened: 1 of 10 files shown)


near2eth/near2eth-block-relay/index.js
```diff
@@ -186,10 +188,53 @@ class Near2EthRelay {
     const maxDelay = Number(near2ethRelayMaxDelay)
     const errorDelay = Number(near2ethRelayErrorDelay)
 
+    const selectDuration = web3.utils.toBN(Number(near2ethRelaySelectDuration) * 1000_000_000)
```
sept-en, line 191: Let's move this variable closer to its usage and also please add time unit suffix in the name

near2eth/near2eth-block-relay/index.js
```diff
@@ -217,8 +262,35 @@ class Near2EthRelay {
...
+              .sub(web3.utils.toBN(lastBlock.inner_lite.timestamp))
+            if (selectDelay.cmpn(0) > 0) {
+              const selectDelaySeconds = selectDelay.div(web3.utils.toBN(1000_000_000))
+              console.log(`Last light client block: ${lastBlock.inner_lite.height}`)
+              console.log(`Time left to make a decision: ${selectDelaySeconds.toString()} seconds`)
+              await sleep(1000) // Block creation time is approximately one second, is an additional cli argument needed here?
+              continue
+            } else {
+              console.log('Time to make a decision is over!')
+            }
```
sept-en, line 290: Let's add some useful logs here (make it as nested logs so it will be clear to read - i.e. starting with "-> ":
* Block size
* Block height

When the block selection is over, print the summary on:
* which block at which height is selected
* how many bytes/gas are saved due to the selection

This will allow showing some stats on how actually this approach is useful.

sept-en, line 290: This logging message is quite useless. When the block is submitted, it's already clear that the decision is made.

near2eth/near2eth-block-relay/index.js
```diff
@@ -217,8 +262,35 @@ class Near2EthRelay {
...
+              nextBlockSelection.set(blockCouple)
+            }
+
+            const selectDelay = selectDuration
+              .add(web3.utils.toBN(nextBlockSelection.startedAt))
+              .sub(web3.utils.toBN(lastBlock.inner_lite.timestamp))
+            if (selectDelay.cmpn(0) > 0) {
+              const selectDelaySeconds = selectDelay.div(web3.utils.toBN(1000_000_000))
+              console.log(`Last light client block: ${lastBlock.inner_lite.height}`)
+              console.log(`Time left to make a decision: ${selectDelaySeconds.toString()} seconds`)
```
sept-en, line 285: Let's make sure that each selection iteration takes 1 log at max. Otherwise, the logs will be flooded with info. Also, please add a verbose option to it. So if verbose is off, only the summary (see my other comment) will be printed when the selection is finished.

Make sure to make messages shorter:
- e.g. `"Time left to make a decision: ${selectDelaySeconds.toString()} seconds"` -> `"${selectDelaySeconds.toString()}s left"`

near2eth/near2eth-block-relay/index.js
```diff
@@ -217,8 +262,35 @@ class Near2EthRelay {
...
+            if (nextBlockSelection.isEmpty()) {
+              console.log('The selection of the optimal block has begun.')
+              nextBlockSelection.set(blockCouple)
+            } else if (nextBlockSelection.isSuitable(blockCouple)) {
+              nextBlockSelection.set(blockCouple)
+            }
+
+            const selectDelay = selectDuration
+              .add(web3.utils.toBN(nextBlockSelection.startedAt))
+              .sub(web3.utils.toBN(lastBlock.inner_lite.timestamp))
```
sept-en, line 281: Add time unit suffix in the name

near2eth/near2eth-block-relay/index.js
```diff
@@ -186,10 +188,53 @@ class Near2EthRelay {
...
+        this.borshBlock = null
+        this.height = 0
+      },
+      isEmpty: function () {
+        return !this.borshBlock
+      },
+      isSuitable: function ({ borshBlock, lightClientBlock }) {
+        return !this.isEmpty() &&
+               this.height !== lightClientBlock.inner_lite.height &&
+               this.borshBlock.length >= borshBlock.length
```
sept-en, line 223: I'd say the `lightClientBlock` is suitable if
```js
this.isEmpty() 
|| (!this.isEmpty()
   && this.height !== lightClientBlock.inner_lite.height
   && this.borshBlock.length >= borshBlock.length)
```

as the block is suitable if there's no more block, isn't it? If we use it, then we can simplify branches with block-selection (when `isEmpty()` and `isSuitable()` are called.

sept-en (changes requested):
Thanks for your PR, this is really helpful!
Added some comments, please address those.

In general, LGTM!

near2eth/near2eth-block-relay/index.js
```diff
@@ -217,8 +262,35 @@ class Near2EthRelay {
...
+            }
+
+            const selectDelay = selectDuration
+              .add(web3.utils.toBN(nextBlockSelection.startedAt))
+              .sub(web3.utils.toBN(lastBlock.inner_lite.timestamp))
+            if (selectDelay.cmpn(0) > 0) {
+              const selectDelaySeconds = selectDelay.div(web3.utils.toBN(1000_000_000))
+              console.log(`Last light client block: ${lastBlock.inner_lite.height}`)
+              console.log(`Time left to make a decision: ${selectDelaySeconds.toString()} seconds`)
+              await sleep(1000) // Block creation time is approximately one second, is an additional cli argument needed here?
```
sept-en, line 286: yes, it would be good to add an argument for this with a default value of 1200 (avg block time now as per explorer info)

sept-en:
@vldmkr please also merge `master` branch and fix merge conflicts

near2eth/near2eth-block-relay/index.js
vldmkr: As I see the `verbose` option is not implemented at the moment. I will take into account other comments and offer a solution with non-flooding logs in the next commit.

near2eth/near2eth-block-relay/index.js
vldmkr: The logic for selecting the suitable block was moved outside the `nextBlockSelection` object for a reason. This is done to preserve the "straightforward" code style of the main loop. This non-obvious decision was commented on [line 271](https://github.com/aurora-is-near/rainbow-bridge/pull/735/files#diff-4aa209577ae540426c39ef90f8b1e1b64b5bd96613ce4f144479bab1b2843a40R271).  I hope there are plans for refactoring :) But I agree. Will be fixed.

karim-en (changes requested):
LGTM, but please just apply the lint fixes.

vldmkr:
Sorry for the delay. Work from Ukraine these days makes its own adjustments.

sept-en (approved):

karim-en (approved):

PR #739, opened 2022-04-20
Updated contract for gas savings and consistency

marc-nyiti:


contracts/eth/nearbridge/contracts/NearBridge.sol
```diff
@@ -269,11 +269,11 @@ contract NearBridge is INearBridge, AdminControlled {
         );
         epoch.numBPs = cnt;
         unchecked {
-            for (uint i = 0; i < cnt; i++) {
+            for (uint i; i < cnt; ++i) {
```
sept-en, line 272: The variable actually should have an initial value explicitly specified.

contracts/eth/nearbridge/contracts/NearBridge.sol
```diff
@@ -269,11 +269,11 @@ contract NearBridge is INearBridge, AdminControlled {
...
                 epoch.keys[i] = src[i].publicKey.k;
             }
             uint256 totalStake = 0; // Sum of uint128, can't be too big.
-            for (uint i = 0; i != cnt; ++i) {
+            for (uint i; i != cnt; ++i) {
```
sept-en, line 276: See my previous comment

contracts/eth/nearbridge/contracts/NearBridge.sol
```diff
@@ -242,7 +242,7 @@ contract NearBridge is INearBridge, AdminControlled {
             untrustedNextHash = nearBlock.next_hash;
 
             uint256 signatureSet = 0;
-            for ((uint i, uint cnt) = (0, thisEpoch.numBPs); i < cnt; i++) {
+            for ((uint i, uint cnt) = (0, thisEpoch.numBPs); i < cnt; ++i) {
```
sept-en, line 245: There's not that much difference between pre- and post- increments here because usually, the compiler is smart enough to optimize this in `for` loops.

sept-en (changes requested):
Hi! Thanks for your PR. Unfortunately, without a clear description, which should be a part of every PR, it's not really clear what you've been trying to achieve. Also, if we are talking about gas savings, please provide some data/tests/stats showing up the results of the optimization.

As for now, closing this as irrelevant, feel free to reopen it if you think this could be useful.

contracts/eth/nearbridge/contracts/NearBridge.sol
marc-nyiti: This is inconsistent throughout your code. In some places you initialize i and other places you leave it uninitialized.  Through my testing leaving i uninitialized and using pre- increments saves a bit of gas.

i initialized

i uninitialized

contracts/eth/nearbridge/contracts/NearBridge.sol
marc-nyiti: I agree there is not much difference between them but you are using both pre- and post- increments throughout your code. better to just stick with one or the other for better consistency.

PR #754, opened 2022-06-14
CI: run Eth-client tests

sept-en:


.github/workflows/contracts-near.yml
```diff
@@ -25,6 +25,13 @@ jobs:
               fi
             timeout-minutes: 40
 
+          - job-name: test eth-client
+            cmd: |
+              cd contracts/near/eth-client
+              ./build.sh
```
karim-en, line 31: Why we call `./build.sh`? 
I think the `./test.sh` should be enough

karim-en (changes requested):

.github/workflows/contracts-near.yml
karim-en, line 31:
```suggestion
```

karim-en (changes requested):

.github/workflows/contracts-near.yml
karim-en: Why do we call `./build.sh`? 
I think the `./test.sh` should be enough

karim-en (approved):

PR #760, opened 2022-07-05
Fix: NEAR proof sharding fix.

karim-en:
* Apply the fixes of `borshifyOutcomeProof` from https://github.com/aurora-is-near/rainbow-bridge-client/commit/555c3c4ee471e6593f29b1b05edb69ed6ea8ccbf
* Remove deprecated `rainbow-bridge-lib` dependency

### DIFF
--- a/contracts/eth/nearprover/package.json
+++ b/contracts/eth/nearprover/package.json
@@ -23,7 +23,6 @@
         "eth-gas-reporter": "^0.2.11",
         "ethers": "5.2.0",
         "hardhat": "^2.1.1",
-        "rainbow-bridge-lib": "^2.0.0",
         "readline-sync": "^1.4.10",
         "solc": "0.8.3",
         "solidity-coverage": "^0.7.16",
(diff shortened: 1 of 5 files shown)


contracts/eth/nearprover/package.json
```diff
@@ -23,7 +23,6 @@
         "eth-gas-reporter": "^0.2.11",
         "ethers": "5.2.0",
         "hardhat": "^2.1.1",
-        "rainbow-bridge-lib": "^2.0.0",
```
sept-en, line 26: Please add `rainbow-bridge-utils` package instead

sept-en (changes requested):

contracts/eth/nearprover/package.json
karim-en: Fixed e3db4be7596a5cf914a1739679065f60e026f130

sept-en (approved):

PR #764, opened 2022-07-31
Eth2OnNearClient: minor improvements, add unit tests 

karim-en:
- Implemented multiple unit tests for eth2 client contract
- Add check for the period update
- Fix bug on unregister submitter account


contracts/near/eth2-client/src/tests/integration_tests.rs
```diff
@@ -0,0 +1,98 @@
...
+            verify_bls_signatures: false,
+            hashes_gc_threshold: 51000,
+            max_submitted_blocks_by_account: 7000,
+            trusted_signer: None,
+        }));
+        let (alice, contract, worker) = initialize_client(init_input).await?;
+
+        alice
+            .call(&worker, contract.id(), "register_submitter")
+            .deposit(10 * ONE_NEAR)
```
olga24912, line 52: Use min_storage_balance_for_submitter?

contracts/near/eth2-client/src/tests/integration_tests.rs
```diff
@@ -0,0 +1,98 @@
...
+            .transact()
+            .await?;
+
+        let num_of_blocks_to_submit = 32;
+        let headers = &headers.as_slice()[1..num_of_blocks_to_submit];
+        for headers_chunk in headers.chunks(50) {
```
olga24912, line 58: I can't understand this code. Are we calling 50 methods in one batch transaction? But before that we take a slice of 32 headers.... And why we take a slice from element number 1. I don't get logic.

contracts/near/eth2-client/src/tests/integration_tests.rs
```diff
@@ -0,0 +1,98 @@
...
+        for header in headers {
+            let result: Option<H256> = contract
+                .view(&worker, "block_hash_safe", header.number.try_to_vec()?)
+                .await?
+                .borsh()?;
+            assert!(result.is_some())
+        }
+        println!(
+            "Gas burnt: {}",
+            gas::to_human(outcome.total_gas_burnt as u128)
```
olga24912, line 94: Will be useful to show more details about how many transactions we done, about bunch and how many headers we finalized, do we send next sync committee or not. 

Also it s good to measure the max number of blocks that we can finalized. And the gas spending dependency of the headers number.

contracts/near/eth2-client/src/tests/integration_tests.rs
karim-en: It is not important for this test

contracts/near/eth2-client/src/tests/integration_tests.rs
karim-en: - The first header in the test data set is the finalized one, so we don't need to submit it.
- I've reduced the `num_of_blocks_to_submit` from 320 to 32 to reduce the time of the ci tests.
- Just edit `num_of_blocks_to_submit` to see how much gas is needed for N blocks
- Yes, the batch size is 50

contracts/near/eth2-client/src/tests/integration_tests.rs
karim-en: We are using the empirical method to see the gas use, do you have ideas for a theoretical solution?
Please note :
- The workspaces tests are slow, so the solution should be optimized as much as possible.
- The validation update was disabled to be able to use the existing test data set, so the gas usage for validating the update should be calculated separately, but it is not important for us because we can see it from logs.

PR #773, opened 2022-08-01
Eth2OnNearClient: change lighthouse dependencies

karim-en:
Change lighthouse dependencies to `wasm-near-wip-downgraded-eth-types` with downgraded `ethereum-types = "0.9.2"`


sept-en (changes requested):
@karim-en I think we need to specify the specific revision to make tracking of changes easier.

sept-en (approved):

PR #795, opened 2022-09-08
Fix `init` for NEAR contracts

karim-en:
* Make init methods private 
* Use `mainnet` feature flag to validate init data


sept-en (changes requested):
@karim-en please fix relayer's test after the changes.

sept-en (approved):

PR #837, opened 2022-10-08
Eth2-to-Near-relay: fix `LightClientUpdate` for new period

olga24912:
We should submit the light client update once in a few epoch to the Eth2 Light Client on Near. 

First, for fetching Light Client Update from Eth2 we used the following API: "eth/v1/beacon/light_client/updates", which fetched the best Light Client Update for specified period. But we faced the problem: the finality block in such Light Client Update did not updated such often as we wanted to. Some time "best" Light Client Update could stay the same for the whole period. 

So we start to use API:  "eth/v1/beacon/light_client/finality_update/", which returns the LightClientUpdate with the last finality block. Unfortunatly such LightClientUpdate does not contain the next sync committee.

Luckily, the next sync committee changing only once per period. So, for fixed this problem we start use the Light Client Update from "eth/v1/beacon/light_client/finality_update/" and manuly included the next_sync_committee from "eth/v1/beacon/light_client/updates". 

However this solution was incorrect. We included not only next_sync_committee, but also the proof of inclusion the next_sync_committee to the finality beacon state. And if we have the another finalized block in the LightClientUpdate we can't just reuse this proof. The test "test_finality_light_client_update_correctness" will fail for the old realization during the checking of the proof for next_sync_committee. 

The correct solution is using "eth/v1/beacon/light_client/finality_update/" for the most of time and using "eth/v1/beacon/light_client/updates" in the beging of new period when we need next_sync_committee. This Light Client Update will be up to date in the begging of the new period.

### DIFF
--- a/eth2near/eth2near-block-relay-rs/src/eth2near_relay.rs
+++ b/eth2near/eth2near-block-relay-rs/src/eth2near_relay.rs
@@ -1,4 +1,3 @@
-use std::cmp;
 use crate::beacon_rpc_client::BeaconRPCClient;
 use crate::config::Config;
 use crate::eth1_rpc_client::Eth1RPCClient;
@@ -19,6 +18,7 @@ use eth_types::eth2::LightClientUpdate;
 use eth_types::BlockHeader;
 use log::{debug, info, trace, warn};
 use near_primitives::views::FinalExecutionStatus;
+use std::cmp;
 use std::cmp::{max, min};
 use std::error::Error;
 use std::thread;
@@ -466,7 +466,8 @@ impl Eth2NearRelay {
         last_finalized_slot_on_eth: u64,
     ) -> bool {
         if (last_submitted_slot as i64) - (last_finalized_slot_on_near as i64)
-            < (ONE_EPOCH_IN_SLOTS * self.interval_between_light_client_updates_submission_in_epochs) as i64
+            < (ONE_EPOCH_IN_SLOTS * self.interval_between_light_client_updates_submission_in_epochs)
+                as i64
         {
             info!(target: "relay", "Light client update were send less then {} epochs ago. Skipping sending light client update", self.interval_between_light_client_updates_submission_in_epochs);
             return false;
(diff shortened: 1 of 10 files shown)


eth2near/eth2near-block-relay-rs/src/eth2near_relay.rs
```diff
@@ -839,6 +840,41 @@ mod tests {
         assert_eq!(finalized_slot, finalized_slot_1);
     }
 
+    #[test]
+    fn test_finality_light_client_update_correctness() {
+        const TREE_FINALITY_DEPTH: usize = 6;
+        const  TREE_FINALITY_INDEX: usize = 41;
```
karim-en, line 846:
```suggestion
        const TREE_FINALITY_INDEX: usize = 41;
```

karim-en: Also, it will be good to move all Ethereum constants to one module including this one 
https://github.com/aurora-is-near/rainbow-bridge/blob/master/eth2near/eth2near-block-relay-rs/src/init_contract.rs#L13

karim-en (changes requested):

eth2near/eth2near-block-relay-rs/src/eth2near_relay.rs
sept-en: @karim-en should be fixed in the scope of a separate PR.

eth2near/eth2near-block-relay-rs/src/eth2near_relay.rs
```diff
@@ -839,6 +840,41 @@ mod tests {
...
+        ));
+
+        let branch = light_client_update.sync_committee_update.as_ref().unwrap().next_sync_committee_branch.iter().map(|h| h.0).collect::<Vec<ethereum_types::H256>>();
+        assert!(merkle_proof::verify_merkle_proof(
+            light_client_update.sync_committee_update.as_ref().unwrap().next_sync_committee.tree_hash_root(),
+            branch.as_slice(),
+            TREE_NEXT_SYNC_COMMITTEE_DEPTH,
+            TREE_NEXT_SYNC_COMMITTEE_INDEX,
+            light_client_update.finality_update.header_update.beacon_header.state_root.0
+        ));
```
sept-en, line 875: Please add messages to `assert` statements. This way, it's better for debugging and also improves readability.

sept-en (changes requested):

eth2near/eth2near-block-relay-rs/src/eth2near_relay.rs
olga24912: fixed: https://github.com/aurora-is-near/rainbow-bridge/pull/837/commits/8cc5447c45673c2c6d0a6094172103918482262a

sept-en (approved):

karim-en (approved):

PR #838, opened 2022-10-12
CI: fix E2E tests

karim-en:
* Update `near-api-js` package to v1.1.0.
* Set `near2eth-relay-block-select-duration` to 0.

### DIFF
--- a/utils/borsh.js
+++ b/utils/borsh.js
@@ -302,7 +302,7 @@ class BorshContract {
           const result = await backoff(10, () =>
             this.account.connection.provider.query({
               request_type: 'call_function',
-              finality: 'optimistic',
+              finality: 'final',
               account_id: this.contractId,
               method_name: d.methodName,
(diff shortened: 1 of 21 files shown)


utils/borsh.js
```diff
@@ -300,10 +300,13 @@ class BorshContract {
...
-              `call/${this.contractId}/${d.methodName}`,
-              nearAPI.utils.serialize.base_encode(args)
-            )
+            this.account.connection.provider.query({
+              request_type: 'call_function',
+              finality: 'optimistic',
+              account_id: this.contractId,
+              method_name: d.methodName,
+              args_base64: args.toString('base64')
+            })
```
sept-en, line 309: Why do we define finality as `optimistic`?

sept-en: Fixed in 91e1ac76239e9818f198427d9aa94dbd308831fb

sept-en (approved):

PR #843, opened 2022-11-06
Eth2Near: minor bug fixes

karim-en:
- Fix beacon rpc route
- Fix network type

### DIFF
--- a/eth2near/eth_rpc_client/src/beacon_rpc_client.rs
+++ b/eth2near/eth_rpc_client/src/beacon_rpc_client.rs
@@ -1,7 +1,10 @@
+use crate::errors::{
+    ErrorOnJsonParse, ExecutionPayloadError, FailOnGettingJson, MissSyncAggregationError,
+    NoBlockForSlotError, SignatureSlotNotFoundError,
+};
 use crate::execution_block_proof::ExecutionBlockProof;
-use crate::errors::{ErrorOnJsonParse, ExecutionPayloadError, FailOnGettingJson, MissSyncAggregationError, NoBlockForSlotError, SignatureSlotNotFoundError};
-use crate::utils;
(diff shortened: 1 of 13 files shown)


sept-en (approved):

eth2near/eth_rpc_client/src/beacon_rpc_client.rs
```diff
@@ -34,7 +34,7 @@ impl BeaconRPCClient {
     const URL_BODY_PATH: &'static str = "eth/v2/beacon/blocks";
     const URL_GET_LIGHT_CLIENT_UPDATE_API: &'static str = "eth/v1/beacon/light_client/updates";
     const URL_FINALITY_LIGHT_CLIENT_UPDATE_PATH: &'static str =
-        "eth/v1/beacon/light_client/finality_update/";
+        "eth/v1/beacon/light_client/finality_update";
```
olga24912, line 37: It can't be change in that way:
`https://lodestar-goerli.chainsafe.io/eth/v1/beacon/light_client/finality_update/`
`https://lodestar-goerli.chainsafe.io/eth/v1/beacon/light_client/finality_update`

For other URLs extra parameters are added and we need not `/`.

olga24912 (changes requested):

eth2near/eth_rpc_client/src/beacon_rpc_client.rs
karim-en: Ok, looks like the route was changed in the last version of lodestar, In this case I will reimplement it by adding versioning for those routes

karim-en: Fixed in 02851a4c8ebe10d0f313423f44440b87a5f09e47

olga24912 (approved):

PR #846, opened 2022-11-20
Change default beacon rpc version

karim-en:


sept-en (changes requested):
Closed as this will be superseded by v1.5.

PR #882, opened 2023-03-29
Eth2Client: improvements, refactoring

karim-en:
Refactor the eth2 light client to resolve an issue of missing finality for more than 5 epochs.

- Added two modes `SubmitLightClientUpdate` and `SubmitHeader`.
- By default, the client is in `SubmitLightClientUpdate` and it turns into `SubmitHeader` after submitting a light client update.
- Added integration tests for migration logic.
- The relayer can submit a `LightClientUpdate` after any number of epochs/blocks, but those can be finalized only if the size of the gap between the current finalized block and the subsequent finalized block is higher than `hashes_gc_threshold`, otherwise, the `GC` can't be applied correctly.
- Update Eth2-to-Near relay and `eth2near` crates to v0.3.0.


sept-en (changes requested):
Having these changes, we also need to add an extra verification to make sure that the chain is convergent. So `block_hash_safe()` method would need to add a check that the block is safe only when `block_number <= finalized_execution_header.block_number`.

But consequently, this also means that `finalized_execution_blocks` variable no longer represents finalized blocks because the only `finalized_execution_header` is actually final.

contracts/near/eth2-client/src/lib.rs
```diff
@@ -259,41 +213,73 @@ impl Eth2Client {
...
-            format!("The block {} already submitted!", &block_hash)
+            format!("The block {:#?} already submitted!", &block_hash)
+        );
+
+        let finalized_execution_header = self.finalized_execution_header.get().unwrap();
+        // Apply gc
+        if let Some(diff_between_unfinalized_head_and_tail) =
+            self.get_diff_between_unfinalized_head_and_tail()
+        {
+            let header_number_to_remove = finalized_execution_header.block_number
```
olga24912, line 246: All of that values is u64. Theoretical (not practical) it is possibility for overflow.

contracts/near/eth2-client/src/lib.rs
```diff
@@ -259,41 +213,73 @@ impl Eth2Client {
...
         require!(
             insert_result.is_none(),
-            format!("The block {} already submitted!", &block_hash)
+            format!("The block {:#?} already submitted!", &block_hash)
+        );
+
+        let finalized_execution_header = self.finalized_execution_header.get().unwrap();
+        // Apply gc
+        if let Some(diff_between_unfinalized_head_and_tail) =
+            self.get_diff_between_unfinalized_head_and_tail()
```
olga24912, line 244: why it is important that it is not None? Maybe better use `unwrap_or(0)`?

olga24912 (approved):

contracts/near/eth2-client/src/lib.rs
karim-en: The`overflow-checks = true` option is enabled

contracts/near/eth2-client/src/lib.rs
karim-en: It is not too important, just to avoid calling the same code twice with `0` on the first and second calls in this mode.

contracts/near/eth2-client/src/lib.rs
```diff
@@ -582,27 +501,11 @@ impl Eth2Client {
...
-
-        self.submitters
-            .insert(submitter, &num_of_submitted_headers.try_into().unwrap());
-    }
-
     fn is_light_client_update_allowed(&self) {
+        require!(matches!(
+            self.client_mode,
+            ClientMode::SubmitLightClientUpdate
+        ));
```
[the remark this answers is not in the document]
karim-en: Fixed 51522a2a361a8807054da731ece3a2b4427e8a60

contracts/near/eth2-client/src/lib.rs
```diff
@@ -54,20 +53,23 @@ pub struct Eth2Client {
...
     /// Submitter account -> Num of submitted headers
-    submitters: LookupMap<AccountId, u32>,
+    __submitters: LookupMap<AccountId, u32>,
     /// Max number of unfinalized blocks allowed to be stored by one submitter account
     /// This value should be at least 32 blocks (1 epoch), but the recommended value is 1024 (32 epochs)
-    max_submitted_blocks_by_account: u32,
+    __max_submitted_blocks_by_account: u32,
     // The minimum balance that should be attached to register a new submitter account
-    min_storage_balance_for_submitter: Balance,
+    __min_storage_balance_for_submitter: Balance,
```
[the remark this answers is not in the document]
karim-en: Removed in 1354ded67703c0799ce076c8c70806af56ca8e30

contracts/near/eth2-client/src/lib.rs
```diff
@@ -259,41 +214,86 @@ impl Eth2Client {
...
+            .as_ref()
+            .map(|header| header.parent_hash)
+            .unwrap_or(self.finalized_beacon_header.execution_block_hash);
+        require!(
+            block_hash == expected_block_hash,
             format!(
-                "Submitted header number {}, hash {:?}",
-                block_header.number, block_hash
+                "The expected block hash is {:#?} but got {:#?}.",
+                expected_block_hash, block_hash
```
olga24912, line 229: Maybe we can also print the block number and expected block number? I think it can be useful in the future for debug.

olga24912 (approved):

contracts/near/eth2-client/src/lib.rs
karim-en: Btw, I've improved the logs a little bit here 
abe2d7b09e5be568d7463210002ed9d823d986d8

sept-en (approved):

PR #896, opened 2023-05-05
`eth-types`: Fix `GasLimitExceeded` error

karim-en:
The hash function `near_sdk::env::keccak256` randomly panics `HostError(GasLimitExceeded)` when calling it not in a near contract because it executes the code in the VM environment.
This code is used by other apps and services that running in the ARM/X86 environments to calculate the block hash, so native sha3 should be used instead of the one from near SDK to improve the performance and to avoid `GasLimitExceeded` error.


olga24912 (approved):

UrAvgDeveloper (commented):
Why can't we use `sha3` package only for both scenarios?

sept-en (changes requested):

sept-en (approved):

PR #898, opened 2023-05-09
CI fix: verify-near-headers, verify-near-proofs

UrAvgDeveloper:
- Fixed data issue for failing CIs - `verify-near-headers` and `verify-near-proofs`
- Updated `nearbridge` tests which were skipped along with header data files used

To fix E2E tests following are the changes implemented:
- Changed ganache version and fixed ganache start script to use automine in some time
- Changed `near-core` version to 1.31.0, because of use of IPv6 port in 1.32.0. Links to verify are [this](https://github.com/near/nearcore/blob/1.31.0/chain/network/src/tcp.rs#L54) and [this](https://github.com/near/nearcore/blob/1.32.0/chain/network/src/tcp.rs#L65). Main issue is GitHub actions not supporting IPv6 ports, the issue for [reference](https://github.com/actions/runner-images/issues/668#ref-commit-5526fae).


contracts/eth/nearbridge/test/NearBridge2.js
```diff
@@ -87,68 +84,64 @@ it('Add second block in first epoch should be verifiable', async function () {
...
-    this.skip();
 
-    const block181 = require('./181.json');
-    const block244 = require('./244.json');
-    const block304 = require('./304.json');
-    const block308 = require('./308.json');
-    const block368 = require('./368.json');
-    const block369 = require('./369.json');
+    const block126313429 = require('./126313429.json');
+    const block126315547 = require('./126315547.json');
```
olga24912, line 89: Should we update the block244 -> block126315547, block304 -> block126315744, block308 -> block126315811 in all tests? Or there is a reason to keep this files different?

contracts/eth/nearbridge/test/NearBridge2.js
```diff
@@ -43,23 +43,20 @@ it('should be ok', async function () {
...
 });
 
-it('2020-09-09 Example', async function () {
-   // Skip until tests are upgraded having blocks after nearcore 1.23.0
-   this.skip();
-
-   const block_15178713 = borshify(require('./block_15178713.json'));
-   const block_15178760 = borshify(require('./block_15178760.json'));
-   const block_15204402 = borshify(require('./block_15204402.json'));
-   const block_15248583 = borshify(require('./block_15248583.json'));
```
olga24912, line 53: We don't need this files anymore and should delete files: "block_15178713.json", "block_15178760.json", "block_15204402.json", "block_15248583.json"

olga24912 (changes requested):

contracts/eth/nearbridge/test/NearBridge2.js
karim-en: The test is about adding blocks from the first epoch, so I think we can keep it for this

contracts/eth/nearbridge/test/NearBridge2.js
karim-en: 1c06fa0cd78cbacb414968d8857501058dc25e8a

olga24912 (approved):

karim-en (approved):

mrLSD (approved):

PR #901, opened 2023-05-11
Update pipeline.yml

Gorlanson:


sept-en (changes requested):
Closing PR as the inappropriate one.

PR #909, opened 2023-07-03
CI: Use the latest `nearcore` version

karim-en:
- Replace `self-hosted` with `ubuntu-latest` to fix the error related to the introduced ipv6 in the [near core  1.32.0](https://github.com/near/nearcore/blob/1.32.0/chain/network/src/tcp.rs#L65)  `thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 99, kind: AddrNotAvailable, message: "Cannot assign requested address" }', chain/network/src/tcp.rs:186:48`

- Revert `utils/scripts/prepare.sh` changes.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -7,7 +7,7 @@ on:
 name: E2E tests
 jobs:
   test:
-    runs-on: self-hosted
+    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


vzctl:
I would like to reduce our dependencies on self-hosted runners, so I approve this change. Let's see if it work well:)

vzctl (approved):

sept-en (approved):

mrLSD (approved):

PR #911, opened 2023-07-14
Cancun upgrade

karim-en:
Handle upcoming Cancun upgrade:

* Add new fields to the block header
```
pub blob_gas_used: Option<u64>,
pub excess_blob_gas: Option<u64>,
pub parent_beacon_block_root: Option<H256>,
```
* Use the workspace version for near contracts.
* Get rid of lighthouse dependencies in the eth-types crate and light client contract.
* Reduce gas usage by removing the `convert_branch` step
* Update lighthouse dependency for the blocks relayer


contracts/near/eth2-utility/src/consensus.rs
```diff
@@ -163,29 +239,41 @@ pub fn get_participant_pubkeys(
...
+        false
+    }
 }
 
-pub fn validate_beacon_block_header_update(header_update: &HeaderUpdate) -> bool {
-    let branch = convert_branch(&header_update.execution_hash_branch);
-    if branch.len() != EXECUTION_PROOF_SIZE {
-        return false;
+/// Compute a root hash from a leaf and a Merkle proof.
+pub fn merkle_root_from_branch(leaf: H256, branch: &[H256], depth: usize, index: usize) -> H256 {
```
olga24912, line 261: I guess we should remove similar function from `eth2near/eth_rpc_client/src/executon_block_proof.rs`

olga24912 (approved):

contracts/near/eth2-utility/src/consensus.rs
karim-en: Yes, but there are different types of conflict issues, so it will be better to solve them in separate PR.

olga24912 (approved):

PR #928, opened 2024-03-15
Add eth proof generation utilities

kiseln:
Replacement for https://github.com/aurora-is-near/rainbow-bridge/blob/master/eth2near/eth2near-block-relay/eth-proof-extractor.js#L86 written in Rust.

Javascript version is used here https://github.com/aurora-is-near/fast-bridge-service/blob/8ed92f1b9c051d49083726803d112aeaa5e51d52/src/ethereum/proof.rs#L40

**Supporting older blocks**
I tested this on post merge and post-4844 blocks. Older blocks may not work (for example because of different handling of difficulty field). If needed I'll test and add support for pre-merge.

**Tests**
I intended to mock RPC requests and add a few tests for the client as well ass more general test cases for proof generation. Would that be the correct approach?

**Error handling**
As one can see it's Box<dyn Error> almost everywhere. I can put some work towards friendlier error types.

**Types**
I chose not to expose the types I added publicly. They are quite specifically designed for deserialization and support only required fields. In the future they can be improved and exposed to other components if that is deemed useful


eth2near/utilities/src/eth_proof_generator.rs
```diff
@@ -0,0 +1,178 @@
...
+
+#[cfg(test)]
+pub mod tests {
+    use super::*;
+    use hasher::Hasher;
+
+    #[test]
+    fn generate_proof() {
+        const RPC_URL: &'static str = "https://eth.llamarpc.com";
+        const TX_HASH: &'static str = "0x9298954a9db8026ca28bce4d71ffb7ba0aac70e91f0667ffb7398c67e60b84fa";
```
karim-en, line 170: Please add more tests for transactions from different epochs(upgrades), e.g. London, Shapella, and Dencun, and compare the results with the proofs generated by the events relayer.

karim-en (changes requested):

eth2near/utilities/src/eth_proof_generator.rs
karim-en: > I intended to mock RPC requests and add a few tests for the client as well ass more general test cases for proof generation. Would that be the correct approach?

For now, it is okay to use the public RPC and make the requests in the tests; we can improve it later in a separate PR by storing the response in the git lfs storage.

karim-en: > Supporting older blocks
> I tested this on post merge and post-4844 blocks. Older blocks may not work (for example because of different handling of difficulty field). If needed I'll test and add support for pre-merge.

Ok, let's ignore pre-merge blocks, and just handle post-merge blocks

eth2near/utilities/src/eth_proof_generator.rs
```diff
@@ -0,0 +1,178 @@
...
+        stream.append(&header.blob_gas_used.unwrap());
+    }
+
+    if header.excess_blob_gas.is_some() {
+        stream.append(&header.excess_blob_gas.unwrap());
+    }
+
+    if header.parent_beacon_block_root.is_some() {
+        stream.append(&header.parent_beacon_block_root.clone().unwrap().0.to_vec());
+    }
```
karim-en, line 155:
```suggestion
    if let Some(base_fee_per_gas) = header.base_fee_per_gas {
        stream.append(&base_fee_per_gas);
    }

    if let Some(withdrawals_root) = header.withdrawals_root.as_ref() {
        stream.append(&withdrawals_root.0.as_slice());
    }

    if let Some(blob_gas_used) = header.blob_gas_used {
        stream.append(&blob_gas_used);
    }

    if let Some(excess_blob_gas) = header.excess_blob_gas {
        stream.append(&excess_blob_gas);
    }

    if let Some(parent_beacon_block_root) = header.parent_beacon_block_root.as_ref() {
        stream.append(&parent_beacon_block_root.0.as_slice());
    }
```

karim-en: We can write some macros to append optional fields to reduce the number of `if` statements. But it is not so important, so let's do that in separate PR.

karim-en (changes requested):

eth2near/utilities/src/eth_proof_generator.rs
```diff
@@ -0,0 +1,178 @@
...
+        receipt_data: encode_receipt(&receipt),
+        header_data: encode_header(&block_header),
+        proof,
+    })
+}
+
+fn build_receipt_trie(receipts: &[TransactionReceipt],) -> Result<PatriciaTrie<MemoryDB, HasherKeccak>, TrieError> {
+    let memdb = Arc::new(MemoryDB::new(true));
+    let hasher = Arc::new(HasherKeccak::new());
+    let mut trie = PatriciaTrie::new(Arc::clone(&memdb), Arc::clone(&hasher));
```
karim-en, line 59:
```suggestion
    let mut trie = PatriciaTrie::new(memdb, hasher);
```
Extra clone

eth2near/utilities/src/eth_proof_generator.rs
```diff
@@ -0,0 +1,178 @@
...
+
+    for (_, receipt) in receipts.iter().enumerate() {
```
karim-en, line 61:
```suggestion
    for receipt in receipts {
```
Extra `enumerate`

eth2near/utilities/src/eth_proof_generator.rs
```diff
@@ -0,0 +1,178 @@
...
+    let block_header = client.get_block_by_number(receipt.block_number)?;
+    let block_receipts = client.get_block_receipts(receipt.block_number)?;
+
+    let trie = build_receipt_trie(&block_receipts)?;
+
+    let receipt_key = rlp::encode(&receipt.transaction_index);
+    let proof = trie.get_proof(&receipt_key)?;
+
+    let mut log_data: Option<Vec<u8>> = None;
+    for (_, log) in receipt.logs.iter().enumerate() {
```
karim-en, line 40:
```suggestion
    for log in &receipt.logs {
```

Extra `enumerate`

eth2near/utilities/src/eth_proof_generator.rs
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    stream.begin_list(4);
+    stream
+        .append(&receipt.status)
+        .append(&receipt.cumulative_gas_used.0)
+        .append(&receipt.logs_bloom.0.to_vec());
+
+    stream.begin_list(receipt.logs.len());
+    for (_, log) in receipt.logs.iter().enumerate() {
```
karim-en, line 85:
```suggestion
    for log in &receipt.logs {
```

eth2near/utilities/src/eth_proof_generator.rs
```diff
@@ -0,0 +1,178 @@
...
+        stream.begin_list(3);
+        stream.append(&log.address.0.to_vec());
+
+        stream.begin_list(log.topics.len());
+        for (_, topic) in log.topics.iter().enumerate() {
```
karim-en, line 90:
```suggestion
        for topic in &log.topics {
```

eth2near/utilities/src/eth_proof_generator.rs
```diff
@@ -0,0 +1,178 @@
...
+    stream.begin_list(4);
+    stream
+        .append(&receipt.status)
+        .append(&receipt.cumulative_gas_used.0)
+        .append(&receipt.logs_bloom.0.to_vec());
+
+    stream.begin_list(receipt.logs.len());
+    for (_, log) in receipt.logs.iter().enumerate() {
+        stream.begin_list(3);
+        stream.append(&log.address.0.to_vec());
```
karim-en, line 87: Please use `as_slice` instead `to_vec` when it possible to avoid extra copy

karim-en (changes requested):

eth2near/utilities/src/eth_proof_generator.rs
```diff
@@ -0,0 +1,178 @@
...
+    pub receipt_data: Vec<u8>,
+    pub header_data: Vec<u8>,
+    pub proof: Vec<Vec<u8>>,
+}
+
+pub fn get_proof_for_event(tx_hash: &str, log_index: u64, node_url: &str,) -> Result<Proof, Box<dyn Error>> {
+    let client = EthRPCClient::new(node_url);
+
+    let tx_bytes: [u8; 32] = hex::decode(&tx_hash[2..])
+        .expect("Invalid hex string")
```
karim-en, line 26: Let's return an error instead of panic here

eth2near/utilities/src/serde.rs
```diff
@@ -0,0 +1,86 @@
...
+use crate::primitives::{Byte, Bytes, FixedBytes};
+
+pub fn hex_to_int_opt<'de, D, N>(deserializer: D) -> Result<Option<N>, D::Error>
+    where D: Deserializer<'de>,
+          N: Num,
+{
+    Ok(Some(hex_to_int(deserializer)?))
+}
+
+pub fn hex_to_int<'de, D, N>(deserializer: D) -> Result<N, D::Error>
```
karim-en, line 12: Some of the ser/de are implemented here https://docs.rs/ethereum_serde_utils/latest/serde_utils/
So please take a look into it and try to reuse them if possible, like here https://github.com/aurora-is-near/rainbow-bridge/blob/6a2e37708ec5a72cc0dc99b7634f0cc796ea9278/contracts/near/eth-types/src/lib.rs#L186-L187

eth2near/utilities/src/eth_proof_generator.rs
kiseln: Fixed in 6a2e37708ec5a72cc0dc99b7634f0cc796ea9278

eth2near/utilities/src/eth_proof_generator.rs
kiseln: Fixed in c70cc73eb4cd78f8dfdfb2f1be34a6807cc5c921

eth2near/utilities/src/eth_proof_generator.rs
kiseln: Fixed in c70cc73eb4cd78f8dfdfb2f1be34a6807cc5c921

eth2near/utilities/src/eth_proof_generator.rs
kiseln: Fixed in c70cc73eb4cd78f8dfdfb2f1be34a6807cc5c921

eth2near/utilities/src/eth_proof_generator.rs
kiseln: Fixed in c70cc73eb4cd78f8dfdfb2f1be34a6807cc5c921

eth2near/utilities/src/eth_proof_generator.rs
kiseln: Fixed in c70cc73eb4cd78f8dfdfb2f1be34a6807cc5c921

eth2near/utilities/src/eth_proof_generator.rs
kiseln: Fixed in c70cc73eb4cd78f8dfdfb2f1be34a6807cc5c921

olga24912 (commented):
Could you please also provide more details in PR description: where this class will be used (why we need this) and the link to JS version 

Also, I think the `utilities` better rename into  `eth_proof_extractor` or something like this.

eth2near/utilities/src/eth_proof_generator.rs
```diff
@@ -0,0 +1,178 @@
...
+    let mut log_data: Option<Vec<u8>> = None;
+    for (_, log) in receipt.logs.iter().enumerate() {
+        if log.log_index == log_index {
+            log_data = Some(encode_log(log));
+        }
+    }
+
+    Ok(Proof {
+        log_index,
+        log_entry_data: log_data.expect("Log not found"),
```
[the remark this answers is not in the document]
kiseln: Fixed in c70cc73eb4cd78f8dfdfb2f1be34a6807cc5c921

eth2near/utilities/src/eth_rpc_client.rs
```diff
@@ -0,0 +1,88 @@
...
+use serde::Deserialize;
+use serde_json::{json, Value};
+use std::error::Error;
+
+pub struct EthRPCClient {
+    endpoint_url: String,
+    client: Client,
+}
+
+impl EthRPCClient {
```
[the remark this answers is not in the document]
kiseln: The reason we decided against using that one is because it depends on types from `contracts/` folder and we wanted to avoid that dependency. Perhaps the new client can replace `eth1_rpc_client` in the future

eth2near/utilities/src/eth_proof_generator.rs
```diff
@@ -0,0 +1,178 @@
...
+    for (_, topic) in log.topics.iter().enumerate() {
+        stream.append(&topic.0.to_vec());
+    }
+
+    stream.append(&log.data.0);
+
+    stream.out().to_vec()
+}
+
+fn encode_header(header: &BlockHeader) -> Vec<u8> {
```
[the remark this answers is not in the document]
kiseln: I think it's because eth-types are under contracts namespace so we avoid using it outside of NEAR smart contracts. Traits could be a cleaner approach I agree

eth2near/utilities/src/eth_proof_generator.rs
```diff
@@ -0,0 +1,178 @@
...
+    pub receipt_index: u64,
+    pub receipt_data: Vec<u8>,
+    pub header_data: Vec<u8>,
+    pub proof: Vec<Vec<u8>>,
+}
+
+pub fn get_proof_for_event(tx_hash: &str, log_index: u64, node_url: &str,) -> Result<Proof, Box<dyn Error>> {
+    let client = EthRPCClient::new(node_url);
+
+    let tx_bytes: [u8; 32] = hex::decode(&tx_hash[2..])
```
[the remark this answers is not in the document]
kiseln: I looked into several type crates and wasn't particularly satisfied but ethereum-types looks good. If @karim-en doesn't object, I can replace our custom types with this library

kiseln:
> Also, I think the `utilities` better rename into `eth_proof_extractor` or something like this.

Depends whether proof generation is the only thing we are going to expose from this crate. If we, say, also want to expose the new ETH client then 'utilities' would work better

eth2near/utilities/src/serde.rs
kiseln: I discovered that this crate doesn't handle odd length strings. If a hex string has leading zeros ETH RPCs send the shortened version.
My code has handling for this (although now I wonder that when there are 2+ leading zeros my code may also fail so I need to handle it)

kiseln: Ethereum spec showing first character cannot be a zero https://github.com/ethereum/execution-apis/blob/main/src/schemas/base-types.yaml#L58

kiseln: Removed some duplication and added tests in 97bc7648291bc25f4e88625f4e589b51c3344209

eth2near/utilities/src/eth_proof_generator.rs
kiseln: Added tests in da40e8ef5a43a364b3e189e9e094293b17da35a4

eth2near/utilities/src/eth_proof_generator.rs
```diff
@@ -0,0 +1,243 @@
...
+    pub header_data: Vec<u8>,
+    pub proof: Vec<Vec<u8>>,
+}
+
+pub fn get_proof_for_event(tx_hash: &str, log_index: u64, node_url: &str,) -> Result<Proof, Box<dyn Error>> {
+    let client = EthRPCClient::new(node_url);
+
+    let tx_bytes: [u8; 32] = hex::decode(&tx_hash[2..])?
+        .try_into()
+        .map_err(|_| Box::<dyn Error>::from("Invalid hex string length"))?;
```
karim-en, line 27:
```suggestion
        .map_err(|_| "Invalid hex string length")?;
```

karim-en (changes requested):

eth2near/utilities/src/eth_proof_generator.rs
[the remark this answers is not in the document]
karim-en: I would like to use `ethereum-types` when possible, but the problem is that these types don't implement Borsh serd.  So let's use it just when possible.

eth2near/utilities/src/eth_proof_generator.rs
```diff
@@ -0,0 +1,243 @@
...
...0000000000000000000000000000010000000010000000000000000000000000000000000000000200000000010000000000000000000000000000000000200080000000200000000000000000000000000000004000000000000002000000000000000000000000000000000000080000000000000000000000000000000000000000000200000004000000000000000000000000000000f901bcf89b94a0b86991c6218b36c1d19d4a2e9eb0ce3606eb48f863a0ddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3efa0000000000000000000000000ef013a60f765f34b0fd7c5aaf83b9c65bb10a9afa000000000000000000000000023ddd3e3692d1861ed57ede224608875809e127fa0000000000000000000000000000000000000000000000000000000174876e800f9011c9423ddd3e3692d1861ed57ede224608875809e127ff863a0dd85dc56b5b4da387bf69c28ec19b1d66e793e0d51b567882fa31dc50bbd32c5a0000000000000000000000000a0b86991c6218b36c1d19d4a2e9eb0ce3606eb48a0000000000000000000000000ef013a60f765f34b0fd7c5aaf83b9c65bb10a9afb8a0000000000000000000000000000000000000000000000000000000174876e8000000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000002f6175726f72613a343332456231663237333036363241443141393739314564333443423244424466373030313535350000000000000000000000000000000000";
```
kisialiou, line 167: I'd propose to store these values in file(s) and load them during testing?

eth2near/utilities/src/eth_proof_generator.rs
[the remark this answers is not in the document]
kiseln: `ethereum-types` has been great. Thanks @olga24912. I managed to replace all custom types except `U8` and `Bytes`

eth2near/utilities/src/eth_proof_generator.rs
kiseln: Thanks, I agree it could be nice. I'm not sure it's worth it, however. There aren't that many of these values and keeping them close to test logic is convenient. Plus, we may want to merge this PR sooner rather than later

olga24912: Also, it will be good to describe how you get this data in the comments. 

I expect, that one day these tests will become out of date, and instruction on how to update the data for tests will save a lot of time in the future. Also, it is another reason to separate test data. So, we can change it without changing the tests itself.

olga24912 (approved):
Thank you for your work! I like the code condition. I haven't tested it yet, but I like how it looks. Thank you for specifying in the PR description the origin code and the place of usage, it makes the reviewing process much easier. 

I continue to think, that it is better to change the name of `utilities` for something more specific. 

It would help if you prepared the PR for https://github.com/aurora-is-near/fast-bridge-service to start using this code instead of JS one.

eth2near/utilities/src/eth_proof_generator.rs
```diff
@@ -0,0 +1,242 @@
...
+
+    let receipt_key = rlp::encode(&receipt.transaction_index);
+    let proof = trie.get_proof(&receipt_key)?;
+
+    let mut log_data: Option<Vec<u8>> = None;
+    let mut log_index_in_receipt = 0;
+    for (i, log) in receipt.logs.iter().enumerate() {
+        if log.log_index == log_index.into() {
+            log_data = Some(encode_log(log));
+            log_index_in_receipt = i;
```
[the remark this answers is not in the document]
kiseln: `log.log_index` is the index within the block

karim-en (approved):

eth2near/utilities/src/eth_proof_generator.rs
kiseln: @kisialiou @olga24912  done in 518d2dae6d00756fdf8885451e61db896aeb802b

PR #944, opened 2025-05-05
Add Sepolia tests for BLS

r-near:
- Added Sepolia testnet data (light client updates + execution blocks)
- New Python script to generate Sepolia test data from live network
- Updated tests to use Sepolia instead of Goerli
- Updated BLS and Ethereum dependencies


contracts/near/eth2-client/src/tests/integration_sepolia.rs
```diff
@@ -0,0 +1,152 @@
...
+    use near_sdk::{Gas, NearToken};
+    use near_workspaces::{Account, Contract};
+    use near_workspaces::operations::Function;
+    use serde::{Deserialize, Serialize};
+
+    const WASM_FILEPATH: &str = "../target/near/eth2_client/eth2_client.wasm";
+    const WASM_V_0_1_0_FILEPATH: &str = "src/data/eth2_client_v0.1.0_testnet.wasm";
+
+    #[derive(Debug, Clone, BorshSerialize, BorshDeserialize, Serialize, Deserialize)]
+    struct BlockHeaderV1 {
```
karim-en, line 19: No need for a separate file for Sepolia tests, the tests should be implemented once for testnet and mainnet, the difference should be just the data

contracts/near/eth2-client/src/data/dump_sepolia_data.py
```diff
@@ -0,0 +1,352 @@
+#!/usr/bin/env python3
+import json
```
karim-en, line 2: The relayer code should be used instead of python script to generate the tests

